PR #12, opened 2026-03-01
Fix non-Latin metadata display for legacy-encoded filenames

hatt-io:
## Summary

- Detects mojibake from legacy Windows codepages (1255/Hebrew, 1256/Arabic, 1251/Cyrillic, 1253/Greek, 874/Thai) misread as Latin-1 and re-decodes to proper UTF-8
- Applies to filename parsing in `TrackFromPath`
- Uses `golang.org/x/text/encoding/charmap` (already an indirect dependency)

## How it works

1. `sanitizeTag()` in new `playlist/encoding.go` counts characters in the Latin-1 supplement range (U+0080–U+00FF) — real Latin text rarely exceeds ~10%, mojibake from non-Latin scripts is almost entirely in this range
2. If above threshold, reverses the Latin-1 decode to recover original bytes
3. Tries UTF-8 first (catches double-encoding), then tries each legacy codepage
4. Picks the encoding that produces the most non-Latin script characters
5. ASCII bytes are identical across all codepages, so mixed-script text is handled correctly

## Test plan

- [x] Files with non-Latin filenames encoded in legacy Windows codepages display correctly
- [x] Files with standard ASCII/UTF-8 filenames are unaffected
- [x] Build: `go build` passes clean

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,6 +9,7 @@ require (
 	github.com/gopxl/beep/v2 v2.1.1
 	github.com/jfreymuth/oggvorbis v1.0.5
 	github.com/madelynnblue/go-dsp v1.0.0
+	golang.org/x/text v0.14.0
 )
 
 require (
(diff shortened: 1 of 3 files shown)


hatt-io:
Sorry about pushing this with conflicts in the first place — should have verified it rebased cleanly before opening the PR.

PR #94, opened 2026-03-16
Fix 403 errors for Spotify playlists saved from other users

### LINKED ISSUE #89: ERR: spotify: list tracks: http status 403 Forbidden: {"error": {"status": 403, "message": "Forbidden" } }
Happens on selecting from my list of playlists after connecting cliamp with spotify using client id

duobei:
## Summary

- `/v1/me/playlists` returns all playlists in the user's library including ones saved/followed from other users. The Spotify API returns 403 when trying to list tracks for these playlists.
- Fetch the current user's ID via `/v1/me` (cached after first call) and filter the playlist list to only show playlists the user owns or collaborates on.
- Add a clear error message for any remaining 403 errors explaining the limitation, instead of showing a raw HTTP error.

## Test plan

- [x] Spotify playlists list no longer shows playlists saved from other users
- [x] Your own playlists and collaborative playlists still appear and load correctly
- [x] Opening a playlist no longer produces `403 Forbidden` errors

Fixes #89

### DIFF
--- a/external/spotify/provider.go
+++ b/external/spotify/provider.go
@@ -38,6 +38,7 @@ type playlistCache struct {
 type SpotifyProvider struct {
 	session    *Session
 	clientID   string
+	userID     string // Spotify user ID, fetched lazily on first Playlists() call
 	mu         sync.Mutex
 	trackCache map[string]*playlistCache // playlist ID → cache entry
 }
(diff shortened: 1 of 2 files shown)


sspaeti:
> Makes sense to add this. I wonder, down the road, if we should look into if you can play playlists that are not yours 🤔

That would be great. I was just confused and got the same error until I read your comment that only my own playlist works (would probably stay somewhere in the docs :) - but the playlists show, and if I hit Enter, I just get an error).

Not sure how hard it is.

PR #103, opened 2026-03-18
Add telemetry opt-out controls and first-run notice

### LINKED ISSUE #102: Telemetry docs mismatch and missing explicit opt-out controls
I noticed a mismatch between the telemetry docs and the actual request, and I think we should tighten the user-facing controls.

Right now `docs/telemetry.md` says a background GET request is sent, but `telemetry/telemetry.go` sends a POST with JSON:

```json
{"uuid":"...","version":"..."}
```

The code path is in `telemetry.Ping()`.

Could we do a small cleanup pass:

- Update `docs/telemetry.md` to match the real request method/payload.
- Add an explicit opt-out in config (for example `telemetry.enabled = false`).
- Add a CLI override (`--no-telemetry`) for one-off runs.
- Show a one-time first-run note that telemetry is enabled and how to disable it.

I’m not arguing against telemetry here. Monthly anonymous MAU is reasonable. This is mostly about trust and making user control obvious.

pythrick:
This PR tightens telemetry transparency and user control.

Changes:
- add config opt-out via `telemetry = false` (also supports `telemetry.enabled = false` and `[telemetry] enabled = false`)
- add CLI session opt-out via `--no-telemetry`
- only call telemetry ping when enabled
- show a one-time first-run notice with disable instructions
- fix telemetry docs to match actual request method/payload (POST JSON)
- document disable options in CLI/config docs
- add tests for config + flag behavior

Fixes #102

### DIFF
--- a/config.toml.example
+++ b/config.toml.example
@@ -13,6 +13,9 @@ shuffle = false
 # Start with mono output (L+R downmix)
 mono = false
 
+# Anonymous monthly telemetry ping (UUID + app version)
+telemetry = true
+
 # Shift+Left/Right seek jump in seconds (6-600)
 seek_large_step_sec = 30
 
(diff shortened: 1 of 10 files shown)


bjarneo:
I appreciate this PR. I wrote the code myself to work the way I want it to!

PR #105, opened 2026-03-18
Refactor resume to session queue model with autosave and takeover lock

### LINKED ISSUE #104: Session resume falls back to sample radio; yt-dlp resume missing
When I close and reopen `cliamp` with no args, it often comes back on the sample radio playlist instead of where I left off.

What I expected:
- reopen to the same playlist
- same track selected
- resume time where possible

What I saw:
- app starts on `https://radio.cliamp.stream/streams.m3u`
- previous playlist is gone for that session

Repro:
1. Start `cliamp` with a playlist (local files or yt-dlp URL).
2. Play a track for a bit.
3. Quit.
4. Start `cliamp` again with no args.

I also hit this with yt-dlp tracks: playlist restore would fail, and position resume did not happen.

I tested a local patch that fixed it by:
- persisting full session state (tracks, current index, shuffle/repeat) on exit
- restoring that state on startup when no positional args are passed
- allowing yt-dlp position resume through the existing seek-by-restart path

If this direction makes sense, I can open a PR.

pythrick:
This replaces the earlier resume implementation with a session queue model and lock-based single-instance handling.

What changed:
- Store the current queue in `~/.config/cliamp/session_queue.toml`
- Store lightweight resume metadata in `~/.config/cliamp/resume.json` (`current_index`, `position_sec`)
- Restore session queue + index + position on startup (when no positional args are passed)
- Autosave session changes with a short debounce and periodic position updates while playing
- Keep shuffle/repeat as user config preferences (not session metadata)
- Add single-instance session lock to prevent concurrent resume file races
- Add `--takeover` to terminate the running instance and acquire the lock

Fixes #104

### DIFF
--- a/config.toml.example
+++ b/config.toml.example
@@ -13,6 +13,9 @@ shuffle = false
 # Start with mono output (L+R downmix)
 mono = false
 
+# Restore last playlist/track/position when reopening with no CLI args
+resume_session = true
+
 # Shift+Left/Right seek jump in seconds (6-600)
 seek_large_step_sec = 30
 
--- a/config/config.go
+++ b/config/config.go
@@ -108,10 +108,10 @@ func (y YouTubeMusicConfig) ResolveCredentials(fallbackFn func() (string, string
 
 // Config holds user preferences loaded from the config file.
 type Config struct {
-	Volume          float64            // dB, range [-30, +6]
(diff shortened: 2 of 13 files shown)


bjarneo:
If this is to be implemented, it has to be done from scratch.

PR #121, opened 2026-03-24
Add RSS auto-discovery for HTML podcast pages

duobei:
## Summary
- When a URL returns `text/html`, automatically discover RSS/Atom feed by parsing `<link rel="alternate" type="application/rss+xml">` in the page head
- Allows playing podcast episodes by pasting the episode page URL directly (e.g. `cliamp 'https://haskell.foundation/podcast/79/'`)
- `sniffFeedURL` now returns the discovered feed URL (may differ from the input URL) instead of a bool

## Test plan
- [x] `go build ./...` passes
- [x] `go test ./...` passes
- [x] Manual: `cliamp 'https://haskell.foundation/podcast/79/'` loads the Buzzsprout RSS feed and plays

### DIFF
--- a/resolve/resolve.go
+++ b/resolve/resolve.go
@@ -183,6 +183,9 @@ func sniffFeedURL(rawURL string) string {
 		return ""
 	}
 	resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		return ""
+	}
 	ct := resp.Header.Get("Content-Type")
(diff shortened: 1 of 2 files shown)


bjarneo:
I appreciate your efforts on this one, but I don't want an approach like this in Cliamp. While I understand the pros of such an approach, it is not often really needed.

PR #170, opened 2026-04-05
Correct keymap scrolling, refactor keymap for consistency

zambetti:
Corrects the scrolling behavior in the keymap view so it matches the persistent scrolling used in the playlist and file browser. Introduces a dedicated `scroll` state for the keymap view, ensuring the view only scrolls when the cursor moves outside the visible region. Additionally, there is some refactoring for style consistency with analogous functions of the file browser view.

### DIFF
--- a/ui/model/keymap.go
+++ b/ui/model/keymap.go
@@ -1,9 +1,13 @@
 package model
 
 import (
+	"fmt"
 	"strings"
 
 	tea "charm.land/bubbletea/v2"
+	"charm.land/lipgloss/v2"
+
+	"cliamp/ui"
 )
 
 // keymapEntry is a key-action pair for the keymap overlay.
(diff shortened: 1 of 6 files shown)


ui/model/keymap.go
```diff
@@ -61,76 +65,152 @@ var keymapEntries = []keymapEntry{
...
 
-	case msg.Code == tea.KeyEscape:
+	case "esc", "ctrl+k":
 		m.keymap.visible = false
-		m.keymap.search = ""
-		m.keymap.filtered = nil
-		m.keymap.cursor = 0
 
-	case msg.Code == tea.KeyUp:
+	case "up", "k":
```
bjarneo: Can you remove the `kjgG`? I would be nice to have them, but then you can't write certain characters while searching the keymap.

PR #172, opened 2026-04-06
feat: support media controller on mac

### LINKED ISSUE #70: Media keys for Mac

gjermundgaraba:
This is an alternative to #171 that takes a slightly broader approach. Instead of adding macOS support via a registry/init() pattern on top of the existing mpris package, this PR consolidates platform media control into a single mediactl package with explicit wiring. 
The mpris code is moved rather than wrapped, dead aliases are removed, and a internal/playback   domain layer is introduced so seek/position values use time.Duration throughout instead of raw microseconds. 
The trade-off is a larger diff, but it eliminates the side-effect import pattern and gives both platforms a clean shared contract.

Side-note: a neat follow-up to this that I am planning is to also get the song/album art showing in the native media controller.

- Add macOS media key support (play/pause/next/prev/seek via Control Centre, Touch Bar, Bluetooth headphones)
- Consolidate MPRIS and macOS backends into a unified mediactl package, replacing the separate mpris package
- Introduce internal/playback domain types using time.Duration instead of raw microseconds
- Use explicit wiring instead of init()-based registry pattern
- Add tests for volume conversion, IPC seek, notifier attachment, and darwin service

image

Closes #70


bjarneo:
I think I like your approach better than mine. I'll spin it up on my mac and give it a go.

PR #190, opened 2026-04-22
Album separators in playlist views

tdimino:
I use playlist view constantly and it was disorienting not seeing where one album ended and another began. This adds visual grouping so multi-album playlists actually read like multi-album playlists.

## Summary

- Album separator lines between album groups in the main playlist view and playlist manager overlay
- Separator-aware scroll calculation—mirrors the existing `providerRowsFromScroll` pattern from the radio list
- Trailing-separator guard: separators only render when there's room for at least one track below
- Docs and site updated

## What changed

3 code files, 2 docs:

| File | Change |
|------|--------|
| `ui/model/scroll.go` | `playlistRowsFromScroll` helper + separator-aware `playlistScroll` |
| `ui/model/view.go` | Separator insertion in `renderPlaylist()` |
| `ui/model/view_overlays.go` | Separator insertion + separator-aware scroll in `renderPlMgrTracks()` |
| `docs/playlists.md` | Note on album grouping |
| `site/index.html` | Playlist feature card updated |

## Test plan
- [x] `make check` passes (IPC socket failures are pre-existing—path length on macOS)
- [ ] `./cliamp --playlist nolan`—separators between Batman Begins, Dark Knight Rises, Inception
- [ ] Scroll through album boundaries—no jitter, cursor stays visible
- [ ] Single-album playlist—one separator at top, none when scrolled
- [ ] Playlist manager (`p`)—separators visible in track list view

## Summary by CodeRabbit

* **New Features**
  * Playlists now group tracks by album and display separator headers in the playlist manager and main player.

* **Bug Fixes**
  * Scrolling and rendering adjusted so the cursor remains visible when album separators are present.
  * Album separators suppressed for streaming tracks to avoid misleading grouping.

* **Documentation**
  * Docs and website copy updated to describe album grouping behavior.

### DIFF
--- a/docs/playlists.md
+++ b/docs/playlists.md
@@ -141,6 +141,8 @@ Press `p` from any view to open the playlist manager:
 6. **Play all**: press `Enter` on the track list to load all tracks into the player
 7. **New playlist**: select "+ New Playlist...", type a name, and press Enter
 
+Tracks with an `album` field are grouped by album with visual separator headers in both the playlist manager and the main player view.
+
 The directory `~/.config/cliamp/playlists/` is created automatically on first use. Removing the last track from a playlist auto-deletes the file.
 
 ### Creating Playlists Manually
--- a/site/index.html
+++ b/site/index.html
@@ -1522,7 +1522,7 @@
       <div class="feature">
         <div class="feature-icon">☰</div>
         <div class="feature-name">Playlists</div>
-        <p>TOML playlists, M3U/M3U8/PLS support, built-in playlist manager.</p>
+        <p>TOML playlists, M3U/M3U8/PLS support, playlist manager with album grouping.</p>
       </div>
       <div class="feature">
         <div class="feature-icon">→</div>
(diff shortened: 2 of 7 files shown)


bjarneo:
Hey

Can you add pictures/video of how it looks?

bjarneo:
It was like this before, but I removed it. Because it might work for the use case you are describing, but the UX does not work for e.g. Spotify playlists. It gets messy.
screenshot-2026-04-23_10-01-06

tdimino:
> Hey
> 
> Can you add pictures/video of how it looks?

It's quite minor, but basically you get a line break between albums in this mode. Left side is my branch; right side is main.

CleanShot 2026-04-23 at 16 54 38@2x

tdimino:
> It was like this before, but I removed it. Because it might work for the use case you are describing, but the UX does not work for e.g. Spotify playlists. It gets messy. screenshot-2026-04-23_10-01-06

Good call on the Spotify case. I added a gate so that album separators now only render for library-style sources (local files, Navidrome, Jellyfin, Plex). Spotify tracks (spotify:track: URIs) skip separators entirely, so playlists with mixed albums stay clean. The check is one function, extensible if other providers hit the same problem.

Issue #206, opened 2026-05-05
Plex integration defaults to first library instead of allowing selection


jkr442:
Cliamp appears to load the first Plex library it finds, which in my case was an audiobook library rather than music. It would be great to either prompt for library selection during setup or let it be defined in the config.

bjarneo:
Feel free to open a PR to improve the Plex implementation!

len0rd:
hey @jkr442  i had a similar issue which i addressed with #211 by adding a new `config.toml` option to select libraries. if that works for you i think this can be closed

PR #211, opened 2026-05-05
plex: add config.toml option for selecting plex libraries

len0rd:
## Summary

Addresses #206. Adds a `libraries` option to the `[plex]` section in `config.toml`. This is a list of libraries to include when displaying/navigating through the Plex provider. This allows users to filter libraries so something like an Audiobook library is not included in the selection

## Screenshots / video

Without filters my personal library looks like this (includes all my audiobooks at the top):

image

Filtering to just include my "music" library ( `libraries = ["Music"]` in config.toml), Audiobooks are no longer shown:

image

## How to test

1. Claude added some unit tests, but otherwise I think you would need a Plex setup with multiple audio libraries to truly test this out
2. I tested this by trying various combinations of the `libraries` option in my config.toml and verifying the expected albums were displayed
   - `libraries` omitted showed both audiobooks and music
   - `libraries = ["Music"]` showed only music
   - `libraries = ["Audiobooks"]` showed only audiobooks
   - `libraries = ["asdfasdf"]` results in an error which directs the user to check their config
   - `libraries = ["Music", "Audiobooks"]` showed both libraries

## Checklist

- [x] `make check` passes
- [x] `docs/` and `site/index.html` updated for user-facing changes

## Summary by CodeRabbit

* **New Features**
  * Optional library filtering for Plex integration — specify library names in config to limit which Plex libraries are loaded.

* **Documentation**
  * Updated Plex configuration docs with examples for the new library filter.

* **Bug Fixes / Behavior**
  * Playlists and library listing now respect the configured library filter; improved messaging when no matching libraries are found.

* **UI**
  * Provider empty-state hint updated to reference library filters in config.

### DIFF
--- a/config.toml.example
+++ b/config.toml.example
@@ -86,6 +86,16 @@ eq = [0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
 # cookies_from = "firefox"   # chrome, brave, edge, opera, safari, vivaldi…
 
 # ---
+# Plex Media Server (optional)
+# [plex]
+# url   = "http://192.168.1.10:32400"
+# token = "xxxxxxxxxxxxxxxxxxxx"
+#
+# Restrict to specific music libraries (comma-separated, case-insensitive).
+# When omitted or empty, all music libraries are loaded.
+# libraries = ["Music", "Jazz"]
+
+# ---
 # Jellyfin server (optional)
 # Authenticate either with a long-lived API token or with your username/password.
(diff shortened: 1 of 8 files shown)


len0rd:
to address the "Misleading error when library filter matches nothing." suggestion, I've slightly tweaked the error message text to say "no **matching** music libraries found...":

image

Coupled with the `Verify [plex] server URL...` line, I think this will make things clearer

bjarneo:
Good addition!

PR #260, opened 2026-06-08
fix(theme): validate and merge user theme files instead of blind override

HANCORE-linux:
## Problem

`loadUserDir()` in `theme/theme.go` overwrites built-in themes with user files from `~/.config/cliamp/themes/*.toml` without any validation. A partial or corrupt file (e.g. only `accent = "#ff0000"`) replaces the entire built-in theme, leaving the other 5 colour fields empty. `lipgloss.Color("")` in `ui/styles.go` then produces broken colours.

Workaround: `rm -rf ~/.config/cliamp/themes/` removes the bad files and restores correct colours -- until the next corrupt file appears.

## Fix

Three new helpers in `theme/theme.go`:

| Function | Purpose |
|----------|---------|
| `validHex(s string) bool` | Validates `#rgb`, `#rrggbb` or `#rrggbbaa` format |
| `Theme.Validate() error` | Checks all 6 colour fields are valid hex |
| `merge(dst *Theme, src Theme)` | Applies only valid-hex fields from `src` onto `dst` |

`loadUserDir` now has two modes:

| Mode | Condition | Behaviour |
|------|-----------|-----------|
| **Merge** | Filename matches built-in (e.g. `dracula.toml` -> `dracula`) | User fields are merged onto the built-in; missing/invalid fields keep the built-in value |
| **Validate** | No built-in match (e.g. `my-theme.toml`) | All 6 hex fields required; file is silently skipped otherwise |

Invalid hex values are silently ignored in either mode -- the corresponding built-in (or zero) value survives. This prevents partial or corrupt files from breaking colours while still allowing single-field customisation.

## Tests

22 tests, all PASS. New tests cover:

- Partial override merging onto a built-in theme
- Full override with all 6 valid fields (merge and validate modes)
- Partial theme without built-in match -> rejected
- Invalid hex field ignored during merge
- `validHex` -- valid: `#fff` `#aabbcc` `#aabbccdd` -- invalid: empty string `#xyz` 2-char hex
- `Validate` with valid, partial, empty, and invalid-hex themes

## Files changed

- `theme/theme.go` -- +validHex, +Validate, +merge, loadUserDir updated
- `theme/theme_test.go` -- +5 tests
- `theme/load_test.go` -- existing tests adapted to merge, +3 new tests

## Summary by CodeRabbit

* **New Features**
  * User themes can partially override built-in themes — specify only colors you want to change while keeping defaults.
  * Stronger theme validation: user themes must provide valid hex color values; invalid entries are ignored.

* **Chores**
  * Improved test coverage for theme discovery, validation, merging, and sorting.
  * User theme lookup now respects a configurable config directory for per-user themes.

### DIFF
--- a/theme/load_test.go
+++ b/theme/load_test.go
@@ -45,19 +45,20 @@ func TestLoadAllSortedCaseInsensitive(t *testing.T) {
 	}
 }
 
-func TestLoadAllUserThemeOverridesBuiltin(t *testing.T) {
+func TestLoadAllPartialUserOverrideMergesOntoBuiltin(t *testing.T) {
 	home := t.TempDir()
 	t.Setenv("HOME", home)
 
-	// Put a user override file named "dracula.toml" with a distinctive accent color.
 	userDir := filepath.Join(home, ".config", "cliamp", "themes")
 	if err := os.MkdirAll(userDir, 0o755); err != nil {
 		t.Fatalf("MkdirAll: %v", err)
 	}
-	overridden := `accent = "#ff00ff"
+
+	// Only two fields — should MERGE onto the built-in dracula.
+	partial := `accent = "#ff00ff"
 fg = "#123456"
 `
-	if err := os.WriteFile(filepath.Join(userDir, "dracula.toml"), []byte(overridden), 0o644); err != nil {
+	if err := os.WriteFile(filepath.Join(userDir, "dracula.toml"), []byte(partial), 0o644); err != nil {
 		t.Fatalf("WriteFile: %v", err)
 	}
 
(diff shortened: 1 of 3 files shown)


HANCORE-linux:
@coderabbitai review

HANCORE-linux:
@coderabbitai review

HANCORE-linux:
Closing this — after testing on a real setup, the validate/merge approach doesn't fix the actual symptom ("wrong / too-bright colours, have to `rm -rf ~/.config/cliamp/themes/`").

The themes that break are **complete, valid** user themes generated by the Omarchy theme-set hook. They pass `Theme.Validate()` and load identically with or without this PR — `cliamp theme list` shows the theme in both the patched and the stock binary, and there's no built-in of the same name, so neither the validate nor the merge path ever engages.

The real bug is the hook's colour mapping, not cliamp:
- `accent = color4` is dark/low-contrast on monochrome themes, yet cliamp uses `accent` for the title, seek bar and key-hint **background** → unreadable.
- `bright_fg`/`fg` (color15/color7) aren't ordered by brightness, so some palettes end up with dimmed text **brighter** than normal text.

`rm` only "helps" because the theme is then not found, `SetTheme` returns false (ignored at startup), and the ANSI defaults apply.

**Real fix** (in the hook's repo): OldJobobo/theme-hook-plugin-manager#3.

Optional cliamp-side hardening (separate, not required for this bug): when `SetTheme(cfg.Theme)` returns false at startup, apply `Default()` and log a warning instead of silently keeping ANSI; and consider not using `accent` as the key-hint background.

Issue #290, opened 2026-08-04
Spotify Windogs


Joshh3ro:
## Spotify Functionality Out Of The Box (Windogs)

So the main issue currently is that windows doesn't support the Spotify setup by default

## Workaround

Id like to enlist some people smarter than me to maybe build a workaround that supports

- Playlists
- Albums
- Tracks

<img width="1919" height="1079" alt="Image" src="https://github.com/user-attachments/assets/561db915-6b2e-44e0-978c-1e31b4cbe46f" />

Joshh3ro:
I dont know if its supposed to be any more detailed but Spotify is sorely missed
I understand we can just download our media but across multiple devices and operating systems

is starting to take a little bit more time than what one might expect especially since i dont want to expose my media server
to the world wide web

Splaxi:
I'm working on fixing the 'go-librespot' for Windows with the support of the maintainer.

https://github.com/devgianlu/go-librespot/issues/365

bjarneo:
> I'm working on fixing the 'go-librespot' for Windows with the support of the maintainer.
> 
> https://github.com/devgianlu/go-librespot/issues/365

That's really awesome!

Splaxi:
What bare min mor features do we need, for it to useful for cliamp?

I have a compilation that works and I was able to use it as a device directly from Spotify client.

So I believe that we are closer than ever 🤞

Splaxi:
Things have been merged 🔥

Can someone pick things up from here? Spent most of my tokens on this one 😬

Joshh3ro:
> I'm working on fixing the 'go-librespot' for Windows with the support of the maintainer.
> 
> [devgianlu/go-librespot#365](https://github.com/devgianlu/go-librespot/issues/365)

That's absolutely amazing ill be keeping an eye out 
And its not "Bare din mor" its also "min mor" 😂😂

Splaxi:
Haha.

Bare minimum features was the real question 😉

I don't know how we move forward from here? Do you need a release with the assembly or will you pull in the repo/code?

Splaxi:
https://github.com/devgianlu/go-librespot/releases/tag/v0.9.0

Joshh3ro:
> https://github.com/devgianlu/go-librespot/releases/tag/v0.9.0

thank you so much ill take a look later, not currently home, but am definitely going to sit down with it once i get a little more time
I'm thinking a release with the assembly, since go and i aren't the best of friends but I'm slowly learning 

[...]

PR #323, opened 2026-08-19
Enable Spotify support on Windows

### LINKED ISSUE #299: Support grouping/sections for custom stations in radios.toml
## Feature request

`~/.config/cliamp/radios.toml` only supports a flat list of `[[station]]`
entries (`name` + `url`). There's no way to group related stations together
in the Radio provider list similarly to `cliamp radio` station, menu.

### Screenshot

Radio provider list

I added 7 stations from one source/country (Kenya) to `radios.toml`, prefixed
with `[🇰🇪 Kenya Radio]` as a naming workaround to keep them visually together.
It works, but it's 7 separate lines cluttering the list where I'd want a
single collapsible entry — similar in weight to the single `cliamp radio`
line above them.

### What I'd like

The Radio provider to support the same "one entry → expands into many"
pattern that Local Playlists already has (a `.toml` playlist shows as one
line under `L`, and Enter reveals its tracks). Something like:

- an optional `group = "Kenya Radio"` key per `[[station]]` entry, rendered
  as a single collapsible group in the Radio provider's list, or
- a nested format, e.g. `[[group]] name = "Kenya Radio"` containing its own
  `[[group.station]]` entries

so a set of related stations can collapse into one line under `R`, the same
way a Local Playlist collapses into one line under `L`.

Happy to help test if this is something you'd consider.

Luguin8:
## Summary

- Windows has been unable to use the Spotify provider because `go-librespot` (the `librespot` playback backend cliamp uses for Spotify) didn't compile there. That was true for `v0.7.1`, which cliamp was pinned to — but `go-librespot` gained first-party WASAPI output for Windows in `v0.9.0` (2026-08-14). This PR bumps the dependency and removes the Windows CGO stub (`external/spotify/stub_windows.go`) so the real provider builds on all platforms.
- CI (`ci.yml`) now installs a MinGW toolchain via `msys2/setup-msys2` and builds/tests with `CGO_ENABLED=1` on the `windows-2025` runner, so this doesn't regress silently. This includes a workaround for an MSYS2 packaging issue: some `mingw-w64-x86_64-libogg` builds ship a `libogg-0.dll` whose export table is missing `ogg_stream_iovecin`, even though it's present in the static `libogg.a` — forcing that one library to link statically (`CGO_LDFLAGS="-Wl,-Bstatic -logg -Wl,-Bdynamic"`) fixes it.
- Updated `README.md` and `docs/spotify.md` to remove the "unavailable on Windows" notes and document the MSYS2/MinGW build steps for Spotify support.
- `session_test.go`: the stored-credentials file-mode assertion (0600) doesn't hold on Windows (NTFS/Go's permission emulation), following the same skip pattern already used in `internal/fileutil/atomic_test.go`.

Fixes #299

## Not in scope for this PR

`release.yml` still builds Windows with `CGO_ENABLED=0`, so the pre-built binaries on the Releases page won't include Spotify until that pipeline is updated separately. Enabling it means either bundling the MSYS2 runtime DLLs (`libvorbis-0.dll`, `libvorbisenc-2.dll`, `libvorbisfile-3.dll`, `libFLAC.dll`, `libmpg123-0.dll`) alongside `cliamp.exe`, or pursuing a fully static build (go-librespot's own `CROSS_COMPILE.md` documents a vcpkg-based static approach) — happy to send a follow-up PR for whichever you'd prefer.

Also found two smaller, unrelated issues while testing — flagging them here rather than folding them into this diff, since they're out of scope:
- `SpotifyProvider.ensureSession()` calls `NewSessionSilent(context.Background(), clientID)` with no timeout. If that hangs (e.g. flaky network), the UI is stuck on "Loading Spotify…" indefinitely with no way to recover short of restarting.
- The `cliamp setup` wizard's "use my own client_id" option didn't let me type in my client_id on Windows (ended up editing `config.toml` by hand instead). Haven't dug into why — could be a real input bug or something environment-specific.

Let me know if you'd like either of those addressed and I can send a PR.

## Test plan

- [x] `go build ./...` — native Windows build, `CGO_ENABLED=1`
- [x] `go vet ./...`
- [x] `go test ./...` — full suite, all green
- [x] Manually verified end-to-end on Windows with a real Spotify Premium account: sign-in, playlist loading, playback, seek, pause/resume

## Summary by CodeRabbit

* **New Features**
  * Added Spotify support for Windows builds, including authentication, playlists, search, and streaming.
  * Released Windows binaries now include Spotify functionality.

* **Bug Fixes**
  * Removed the Windows-only “Spotify unavailable” behavior and messaging.

* **Documentation**
  * Added Windows setup instructions covering required build tools, libraries, and runtime files.

* **Tests**
  * Expanded Spotify build and test coverage to include Windows.


bjarneo:
Connecting issue to PR: https://github.com/bjarneo/cliamp/issues/290

Merging. Noted that it will still built without.

PR #331, opened 2026-08-20
fix(seek): debounce seeks that restart a decoder

coryshaw1:
## Summary

Holding or tapping the seek keys on a buffered network stream freezes the UI. Each press runs `Player.Seek` inline on the UI goroutine, and for these streams a seek kills ffmpeg and restarts it with `-ss`, then waits for PCM — so three quick presses serialise three restarts and the interface stops responding until they finish.

`seekRelative` already had the fix for yt-dlp: presses accumulate into `seek.targetPos` and one seek fires ~800ms after the last, from a command goroutine. This routes decoder-restarting stream seeks through that same path rather than adding a second mechanism.

`needsDebouncedSeek` gates it on `track.Stream && Seekable()`, which covers the providers sharing the nav/ffmpeg pipeline — Navidrome, Plex, Jellyfin, Emby, Qobuz, Audiobookshelf. Local files, radio (not seekable), Spotify (`Stream: false`, and its decoder repositions in place), and NetEase keep seeking inline.

Worth noting this mitigates rather than removes the cost: ffmpeg reads these streams through `pipe:0`, so it cannot seek and has to restart, even though `navBuffer` already has the bytes on disk. Feeding it the temp file instead would make seeks cheap and this debounce unnecessary — a larger change, and one I would rather raise separately than bundle here.

## How to test

1. Play an Audiobookshelf (or Navidrome/Plex/Jellyfin/Emby/Qobuz) track and tap the right arrow three or four times quickly. The UI should stay responsive, and one seek should land ~800ms after the last press at the summed offset.
2. Try `shift`+arrow for the larger step, and the left arrow to confirm backward seeks sum the same way.
3. Seek a local file and confirm it still moves immediately, with no added delay.
4. Seek a yt-dlp source (`cliamp "https://www.youtube.com/watch?v=<id>"`) and confirm its existing debounce still works — this PR renames the functions it uses.
5. Play a radio stream and confirm the arrow keys behave as before.

## Checklist

- [x] `make check` passes

No docs change: nothing in `docs/` or `site/index.html` describes seek timing, so there is no stale claim to correct.

I tested Audiobookshelf, yt-dlp, Spotify, radio and local files by hand. Navidrome, Plex, Jellyfin, Emby and Qobuz share the same decoder and predicate so should behave identically, but I have no server for those and have not verified them directly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved seeking responsiveness by combining rapid seek requests for supported playback and stream tracks.
  * Preserved immediate seeking when decoder restarts are unnecessary.
  * Improved handling of rapid, timer-triggered, absolute, and resume seek actions.
  * Ensured pending seek requests apply correctly after an active seek completes.
  * Prevented outdated seek results from affecting the current track.
  * Improved resume behavior during Mixcloud playback.


bjarneo:
Can you fix the merge conflict?

PR #330, opened 2026-08-20
feat: cross-playlist favorites virtual playlist

tahadx:
## Summary

Adds a cross-playlist favorites system that lets users mark tracks with ♥ from any playlist and collects them into a persistent virtual **Favorites** playlist.

> **Connected to #329** — this branch is stacked directly on top of `feat/dir-source-tui-manager` and was rebased onto its latest commit. It uses the new playlist-manager flow from #329 (create/rename/delete, `[[dir]]` sources) and extends it with Favorites guards. **Merge #329 first**; this diff shrinks to the favorites-only changes once it lands.

## What changed

- **`n` key** toggles favorite on any track (main view and playlist manager tracks screen)
- **Favorites virtual playlist** always appears at the top of the playlist list (even when empty, for discoverability)
- **♥ count** `[♥ N]` in the playlist header shows total favorites
- **♥ marker** on favorited tracks in the track list (uses cached `favSet` map — no per-frame disk I/O)
- **`f/n ★/♥` hint** in the help bar
- **Manager integration (new)** — Favorites gets the same virtual-playlist protections as Recently Played: `r`/`d`/`D` show friendly notices instead of raw save errors, and `d` on a Favorites track points at `n`

## How it works

- New `favorites/` package — TOML persistence at `~/.config/cliamp/favorites.toml`
- `provider.FavoritesManager` interface implemented by `external/local`
- `favSet` map on UI model for O(1) render lookups
- Separate from bookmarks (`f` key, ★ marker) — bookmarks are per-playlist, favorites span all playlists

## Tests

- `favorites/favorites_test.go` — Store lifecycle, TOML round-trip, edge cases
- `external/local/provider_test.go` — Favorites in Playlists(), Tracks(), ToggleFavorite(), FavoritesCount()
- `ui/model/dirs_screen_test.go` — N key toggles favorite, no-op without favMgr, Favorites rename/delete/dir-source guards
- `cmd/playlist_ops_test.go` — PlaylistList shows Favorites even when empty

PR #333, opened 2026-08-21
feat: cross-playlist favorites virtual playlist

tahadx:
Replaces closed #330 (its branch was force-pushed during the rebase, so GitHub blocked reopening).

## Summary

Adds a cross-playlist favorites system that lets users mark tracks with ♥ from any playlist and collects them into a persistent virtual **Favorites** playlist.

> **Connected to #329** — this branch is stacked directly on top of `feat/dir-source-tui-manager` and was rebased onto its latest commit. It uses the new playlist-manager flow from #329 (create/rename/delete, `[[dir]]` sources) and extends it with Favorites guards. **Merge #329 first**; this diff shrinks to the favorites-only changes once it lands.

## What changed

- **`n` key** toggles favorite on any track (main view and playlist manager tracks screen)
- **Favorites virtual playlist** always appears at the top of the playlist list (even when empty, for discoverability)
- **♥ count** `[♥ N]` in the playlist header shows total favorites
- **♥ marker** on favorited tracks in the track list (uses cached `favSet` map — no per-frame disk I/O)
- **`f/n ★/♥` hint** in the help bar
- **Manager integration** — Favorites gets the same virtual-playlist protections as Recently Played: `r`/`d`/`D` show friendly notices instead of raw save errors, and `d` on a Favorites track points at `n`
- **Live counts** — unfavoriting inside the Favorites screen removes the row immediately; the Favorites label always shows its count (`Favorites · 0 tracks` when empty); every toggle re-pulls the manager list and Local pane so counts never go stale

## How it works

- New `favorites/` package — TOML persistence at `~/.config/cliamp/favorites.toml`
- `provider.FavoritesManager` interface implemented by `external/local`
- `favSet` map on UI model for O(1) render lookups
- Separate from bookmarks (`f` key, ★ marker) — bookmarks are per-playlist, favorites span all playlists

## Tests

- `favorites/favorites_test.go` — Store lifecycle, TOML round-trip, edge cases
- `external/local/provider_test.go` — Favorites in Playlists(), Tracks(), ToggleFavorite(), FavoritesCount()
- `ui/model/dirs_screen_test.go` — N key toggles favorite, no-op without favMgr, Favorites rename/delete/dir-source guards, row removal on unfavorite, count refresh after toggles from queue and manager
- `cmd/playlist_ops_test.go` — PlaylistList shows Favorites even when empty

## Summary by CodeRabbit

* **New Features**
  * Added a virtual **Favorites** playlist with cross-playlist favorite toggling, counts, persistence, and visual markers.
  * Added directory sources for playlists, including recursive scanning, removal, and management controls.
  * Expanded playlist creation and file-browser workflows for adding files and directories.
  * Added keyboard shortcuts for favorites, playlist management, directory sources, and provider access.

* **Bug Fixes**
  * Favorites now appears even when empty.
  * Playlist metadata and displayed counts refresh after changes.

* **Documentation**
  * Updated keyboard shortcut, playlist, and feature documentation.


bjarneo:
I used AI to help review this change. I found the following possible issues. Please verify each issue.

Some issues are in the changes from #329 that are included in this PR.

1. **An existing `Favorites.toml` playlist can become inaccessible.**

   `Playlists()` can list a physical playlist named `Favorites`. `Tracks("Favorites")` always opens the virtual Favorites store. The save, rename, and delete operations also reject this name. Thus, a user cannot access or manage the physical playlist. Please add a migration to a safe name before you reserve `Favorites`.

   Reference: [`external/local/provider.go:94-132`](https://github.com/bjarneo/cliamp/blob/c8c3d33358b331fe877b8ade4dc07f34219eab3a/external/local/provider.go#L94-L132)

2. **Some non-local favorites do not work correctly after a restart.**

   The Favorites file does not store `Realtime`, `ProviderMeta`, and other playback data. A radio favorite loses its live-stream behavior. Navidrome and Jellyfin favorites lose playback reports. Qobuz stores a signed URL that can expire. Please store a stable provider identity and resolve a new playback URL when necessary.

   Reference: [`favorites/favorites.go:267-355`](https://github.com/bjarneo/cliamp/blob/c8c3d33358b331fe877b8ade4dc07f34219eab3a/favorites/favorites.go#L267-L355)

3. **Undo can lose directory sources after playlist deletion.**

   This issue is in the #329 stack. The delete operation stores only the expanded tracks. It does not store the `[[dir]]` sections. Undo creates a new file and removes tracks that have `DirSourced=true`. The result can be an empty or incomplete playlist, although the UI reports success. Please store and restore the complete playlist document.

   References: [`ui/model/keys.go:1655-1658`](https://github.com/bjarneo/cliamp/blob/c8c3d33358b331fe877b8ade4dc07f34219eab3a/ui/model/keys.go#L1655-L1658), [`external/local/provider.go:621-643`](https://github.com/bjarneo/cliamp/blob/c8c3d33358b331fe877b8ade4dc07f34219eab3a/external/local/provider.go#L621-L643)

4. **The Windows file lock does not lock the file.**

   `LockFile` only opens the lock file on Windows. Two cliamp processes can read the same Favorites data. Each process can then write a different result. The last write removes the update from the other process. The TUI can continue after an IPC startup error, so two processes are possible. Please use a Windows lock such as `LockFileEx`. Please add a test that verifies lock contention.

   Reference: [`internal/fileutil/lock_windows.go:10-23`](https://github.com/bjarneo/cliamp/blob/c8c3d33358b331fe877b8ade4dc07f34219eab3a/internal/fileutil/lock_windows.go#L10-L23)

5. **A late file-browser result can recreate a deleted or renamed playlist.**

   This issue is in the #329 stack. The asynchronous result stores only the target name. It has no request generation or existence check. If the user deletes or renames the playlist before resolution finishes, `AddTracks` creates the old playlist again. Please reject stale results or verify the playlist identity before the write.

   References: [`ui/model/filebrowser.go:532-555`](https://github.com/bjarneo/cliamp/blob/c8c3d33358b331fe877b8ade4dc07f34219eab3a/ui/model/filebrowser.go#L532-L555), [`ui/model/update.go:614-633`](https://github.com/bjarneo/cliamp/blob/c8c3d33358b331fe877b8ade4dc07f34219eab3a/ui/model/update.go#L614-L633)

6. **A failed playback attempt can enter Recently Played.**

   This issue is in the #329 stack. `beginPlaybackTrack` records history before local playback succeeds. It also records history before asynchronous stream playback succeeds. A corrupt file, an expired URL, or a restricted stream can appear in Recently Played. Please record the track only after playback starts successfully.

   Reference: [`ui/model/playback.go:441-452`](https://github.com/bjarneo/cliamp/blob/c8c3d33358b331fe877b8ade4dc07f34219eab3a/ui/model/playback.go#L441-L452)

7. **The UI still shows Favorites as a writable playlist.**

   The write picker and the search playlist picker remove Recently Played, but they do not remove Favorites. Favorites is usually the first item. A write to it always fails. The manager also permits file add, sort, and reorder actions for virtual playlists. Please remove all virtual playlists from write destinations. Please disable all unsupported actions.

   References: [`ui/model/pl_picker.go:16-31`](https://github.com/bjarneo/cliamp/blob/c8c3d33358b331fe877b8ade4dc07f34219eab3a/ui/model/pl_picker.go#L16-L31), [`ui/model/commands.go:490-500`](https://github.com/bjarneo/cliamp/blob/c8c3d33358b331fe877b8ade4dc07f34219eab3a/ui/model/commands.go#L490-L500)

8. **Concurrent history writers can lose entries.**

   This issue is in the #329 stack. The mutex protects only one `Store` instance. An atomic rename prevents a damaged file, but it does not protect the complete read-modify-write operation. Two processes can read the same history. The last process to write removes the new entry from the other process. Please use an interprocess lock for `Record` and `Clear`.

   Reference: [`history/history.go:86-113`](https://github.com/bjarneo/cliamp/blob/c8c3d33358b331fe877b8ade4dc07f34219eab3a/history/history.go#L86-L113)

9. **Local writes can refresh the active remote provider.**

   Favorite changes and history changes call `fetchProviderPlaylists()` directly. If Spotify or Navidrome is active, cliamp starts an unrelated remote request. This request can replace another request or show a remote error. Please use `refreshPaneAfterLocalWrite()` for local changes.

tahadx (commented):
**Reviewer disclosure:** this review was written by an AI coding assistant (ox-alpha) at the contributor's request, and posted with their approval. Please weigh it alongside human review. Note this diff includes the stacked #329 changes; I focused on the favorites-only delta as flagged in the body.

## What I checked

The favorites store is built on the right foundations: a per-instance mutex plus an advisory `flock` (`internal/fileutil.LockFile`) serializes load-modify-save across cliamp processes, writes go through atomic temp-file+rename with fsync, `parse` ignores unknown keys for forward compatibility, and entries that fail to parse a path are dropped rather than corrupting the list. The `favSet` map giving O(1) render lookups (no per-frame disk reads) is exactly how to do the ♥ marker, and `refreshFavSet` centralizes invalidation after each toggle. On the manager side, `plMgrVirtualPlaylistName` centralizing the Favorites/Recently-Played protections is a genuine improvement over scattering name comparisons — rename/delete/dir-source guards all route through it, and `d` on a Favorites track pointing at `n` instead of erroring is friendly UX.

## Observations

- **Windows lock is a documented no-op:** `lock_windows.go` opens the lock file but takes no lock, so multi-process Windows builds can still race on favorites.toml. The code comment says this honestly, and single-process desktop use makes it low-risk — just noting it's best-effort there by design.
- **Synchronous toggle on the UI thread:** `ToggleFavorite` blocks on the flock. That's safe in practice — writes are sub-millisecond and `flock` auto-releases when a process dies — but it means UI latency depends on another process's write window. Fine as-is; flagging only so nobody adds heavier work under the same lock later.
- **Path-keyed identity:** favorites match on exact path, so a moved/renamed file silently orphans its entry. Reasonable for v1; worth one line in the docs if users ever ask where a favorite went.
- **Per-toggle `Playlists()` pulls:** like `recordListenedTrack` in #329, the `n` handler returns `m.fetchProviderPlaylists()` unconditionally, which re-parses every playlist TOML and walks every `[[dir]]` source per keypress. Gating on Local being active (the pane can't show Local counts otherwise) would keep counts fresh at zero cost when browsing other providers.
- **Favorites always visible even when empty** is a deliberate discoverability choice per the body — reasonable, though it does add a permanent row; the `· 0 tracks` label keeps it honest.

Tests cover store lifecycle/TOML round-trip, provider interface compile checks, both `n` flows including the no-favMgr no-op, count refresh from the queue, and scroll reset on empty reload — thorough coverage of the seams most likely to regress.

Stacking discipline is also clean: merge order documented, guards shared with #329's work, and the diff genuinely shrinks once the base lands.

bjarneo:
Can you resolve the conflicts

tahadx:
Of course!
also fixed bugs

PR #335, opened 2026-08-21
feat(ui): mouse support for playlist, providers, seek and volume

coff33ninja:
## Summary
- Adds phase-1 mouse support to the TUI: click-to-select playlist/provider rows (clicking an already-selected row plays it, mirroring Enter), seek-bar and volume-bar click handling, and wheel scrolling that matches keyboard navigation semantics (clamped, no wrap; provider wheel-down loads catalog batches).
- Includes `scripts/build-zig.ps1`, a reproducible Windows cgo builder that uses zig as the only C toolchain.

## How it works
- Renderers record clickable regions into a per-frame scratch state (`mouseHitState`, pointer field on `Model` like `pluginEmit`).
- `mainSections` tracks the running line offset of each section and resolves scratch regions into absolute frame coordinates; the visualizer's centered frame records its offsets so clicks inside it map correctly.
- `View()` enables `tea.MouseModeCellMotion`; `Update` dispatches `tea.MouseClickMsg` / `tea.MouseWheelMsg` with the same post-processing as key presses.
- Seek clicks require a known duration and are ignored while buffering, matching keyboard seek behavior. Volume maps continuously over [-50, +6] dB.

## Builder script notes (Windows cgo without MinGW)
- MSYS2 libogg packages ship with IOVEC disabled, but xlab/vorbis-go references `ogg_stream_iovecin` at link time, so the script compiles libogg 1.3.6 from source instead.
- `libFLAC.dll` depends on `libwinpthread-1.dll`, which current MSYS2 winpthreads packages no longer ship as a DLL; the script builds it from zig's bundled winpthreads sources.
- A minimal pkg-config shim answers only `--cflags`/`--libs`; everything is vendored under gitignored `.cgo-deps/`.

## Testing
- New `ui/model/mouse_test.go`: MouseMode enabled, select-then-play, clamped wheel scrolling, seek-to-position accuracy (~30min target +-5s on a 60-min track), volume endpoints (-50 exact / +6 computed), no row hits while overlays are open, Update dispatch wiring, provider search-filter index resolution.
- Full `go test ./...` passes, including under the zig/cgo environment produced by `scripts/build-zig.ps1 -Test`.

## Summary by CodeRabbit

* **New Features**
  * Added mouse interaction support for selecting and playing playlist tracks.
  * Added clickable seek and volume controls.
  * Added mouse-wheel scrolling for playlists and provider lists.
  * Added interactive provider playlist and catalog loading.
  * Added a reproducible Windows build setup through PowerShell.

* **Bug Fixes**
  * Prevented scrolling beyond available playlist and provider content.
  * Suppressed background row interactions while overlays are open.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@ docs/ideas.md
 config.backup.toml
 .env
 .github/workflows/contextowl-changelog.yml
+.cgo-deps/
+/zig-build/
--- a/scripts/build-zig.ps1
+++ b/scripts/build-zig.ps1
@@ -1,0 +1,264 @@
+#Requires -Version 7.0
+<#
+.SYNOPSIS
+    Reproducible Windows cgo build of cliamp using only zig as the C toolchain.
+
+.DESCRIPTION
+    Provisions a self-contained toolchain under .cgo-deps/ (vendored MSYS2 dev
+    libraries, a minimal pkg-config shim, a source-built libogg with IOVEC
+    support, and libwinpthread-1.dll compiled from zig's bundled winpthreads),
+    then builds the binary and copies the required runtime DLLs next to it.
+
+.PARAMETER Output
+    Path of the built executable. Defaults to <repo>\zig-build\cliamp.exe.
+    Runtime DLLs are copied next to it.
+
+.PARAMETER Test
+    Run the full test suite (go test ./...) in the same environment after building.
+
+.PARAMETER ProvisionOnly
+    Provision .cgo-deps/ without building.
+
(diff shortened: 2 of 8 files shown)


bjarneo:
As this is a tui, do we need mouse support?

theandrewbailey:
> As this is a tui, do we need mouse support?

htop supports mouse features, despite being TUI. I would appreciate it in cliamp. Might open the door to smartphone ports too.

davidszp:
Track seeking would see a large benefit from mouse support

PR #340, opened 2026-08-22
fix(player): decode live streams off the speaker mutex

### LINKED ISSUE #339: Live radio stream freezes UI and audio during long sessions (speaker mutex held on slow network reads)
## Summary

During long live-radio playback sessions, the app periodically freezes both audio and UI (seek bar, navigation, any keypress) for anywhere from ~100ms to over a second. Frequency increases the longer the stream stays connected. Not specific to one radio station/server — reproduced against a plain MP3 Icecast stream.

## Root cause

`beep`'s `speaker` package holds a single global mutex (`speaker.mu`) for the duration of every `Stream()` call on the registered streamer (`gopxl/beep/v2/speaker.sampleReader.stream`). For live (non-seekable) HTTP sources, that `Stream()` call chain ends in a native decoder (`mp3.Decode`, etc.) doing a synchronous, blocking `Read()` on the HTTP response body wrapped by `stallReader` (`player/decode.go`).

`streamStallTimeout` (10s) only guards against a fully dead/half-open connection. It does nothing for a live-but-degraded connection that keeps delivering data, just slower than real-time playback requires — small chunks arriving 100ms-900ms+ apart is common on long-lived Icecast/CDN connections and never approaches the 10s threshold.

Every UI tick calls `Player.PositionAndDuration()`, which takes `speaker.Lock()` before reading position. So each slow network `Read()` inside the audio callback blocks every other `speaker.Lock()` caller (position polling, pause, volume, seek) for the same duration — the whole TUI appears to hang, not just the audio.

## Reproduction / evidence

Instrumented `stallReader.Read()` and `PositionAndDuration()` with debug logging against a live production stream over several hours. Sample from the log, one 1:1-paired incident:

```
stallReader: slow Read took 761.611041ms (n=888 err=<nil>)
PositionAndDuration: speaker.Lock() waited 762.203363ms
```

This pattern repeats continuously during affected sessions, at varying magnitudes (observed 100ms to ~1s), always paired with a same-duration UI mutex wait. Confirmed it is *not* momentary jitter: increasing the read buffer ahead of the decoder (tried a 2MB ring buffer) did not reduce frequency — the bottleneck is sustained network throughput below real-time playback rate for that connection, not a brief stall, so any finite buffer still drains to empty and blocks again.

Ruled out before isolating this: disk I/O pressure, NVMe controller errors (real, fixed separately, but unrelated), CPU contention from other running apps, and a stallReader network-decoupling attempt (moves the blocking Read to a background goroutine — helps, but the *consumer* side of that buffer still blocks inside `Stream()` when it drains, so the UI still freezes on sustained slow throughput, just slightly less often).

## Fix

Opening a PR with a fix: a `livePrefetchStreamer` that runs the actual decode chain in a dedicated goroutine against a ring buffer of already-decoded samples. The `Stream()` method registered with `speaker` only ever copies from that buffer and never blocks — on underrun it returns silence for the requested length instead of waiting. This trades a same audio glitch (which is unavoidable — the data genuinely hasn't arrived) for never holding `speaker.Lock()` while waiting on the network. Wired in only for live/non-seekable HTTP sources; buffered/seekable sources (Navidrome etc.) already get equivalent decoupling from `navBuffer`'s background download goroutine.

## Environment

- cliamp v1.63.2 (also reproduced on current `main` / v1.63.2+ with the `perf: optimize playback` commit)
- Arch Linux, PipeWire
- Stream: plain MP3 Icecast (`Content-Type: audio/mpeg`, no `Icy-Metaint` involved in reproduction)

OtavioPompolini:
Fixes #339

## Problem

Live (non-seekable) HTTP streams — internet radio — decode synchronously inside the beep speaker's audio callback, which holds `speaker`'s global mutex for the duration of the call. When the network delivers data slower than real-time playback needs (small, delayed chunks — common on long-lived Icecast/CDN connections, well under `streamStallTimeout`), that blocking `Read()` holds the mutex for the same duration. Every UI operation that needs it (`PositionAndDuration()` on every tick, pause, volume, seek) freezes along with the audio. See #339 for the full root-cause writeup and log evidence from a reproduced session.

## Fix

`livePrefetchStreamer` (new, `player/live_prefetch.go`) wraps the decode chain for live/non-seekable HTTP sources. It runs the actual `Stream()` calls in a dedicated goroutine, writing finished samples into a ring buffer sized for a few seconds of audio. The `Stream()` method actually registered with `speaker` only ever copies from that buffer and never blocks: on underrun (buffer empty, source not keeping up) it returns silence for the requested length rather than waiting.

This trades a brief audio glitch for never holding `speaker.Lock()` on network I/O — the glitch is unavoidable (the data genuinely hasn't arrived yet), but the UI no longer freezes with it.

Wired in only where it's needed: `buildPipeline` wraps the stream with `livePrefetchStreamer` when `!seekable` (the existing flag that's already true only for live HTTP sources). Buffered/seekable sources (Navidrome, etc.) are untouched — `navBuffer` already gives them equivalent decoupling via its own background download goroutine.

`trackPipeline.close()` closes the prefetcher to stop its fill goroutine, mirroring how `decoder.Close()` is already handled.

## Testing

- Added `player/live_prefetch_test.go`: never-blocks-under-stall, real samples delivered once decoded, EOF propagation after drain, `Err()` propagation, and `Close()` stopping the fill goroutine promptly.
- `go test ./...` and `go test -race ./player/...` both pass.
- `go vet ./...` and `gofmt -l .` clean.
- Manually verified against the live production stream used to diagnose the issue: before the fix, `speaker.Lock()` wait times matched network read stall durations 1:1 (up to ~900ms) with visible UI/audio freezes; after, the same session ran without freezes (instrumented debug logging removed from this PR — this was a throwaway verification build).

## Scope note

Not claiming this eliminates every possible glitch on a sustained-too-slow connection — that's a real network/server limitation and some audio interruption is unavoidable there. This fixes the part that shouldn't happen regardless: the UI (and by extension all playback control) freezing along with it.

## Summary by CodeRabbit

* **Documentation**
  * Clarified audio playback behavior for stalled requests and delayed live-stream data.
  * Added context around live-stream prefetching, including how it applies to non-seekable streams and is stopped during shutdown.
* **No User-Facing Changes**
  * These updates improve technical clarity without changing playback functionality.

### DIFF
--- a/player/decode.go
+++ b/player/decode.go
@@ -207,9 +207,12 @@ func openSource(path string, onMeta func(string)) (sourceResult, error) {
 	}
 
 	// Guard the body with a stall timeout so a stalled/half-open live stream
-	// cannot park a Read forever. Slow reads are moved off the speaker callback
-	// by livePrefetchStreamer; this timeout handles a connection that stops
-	// making progress entirely.
+	// can't park a Read forever. Close() cancels the request, so this also
+	// cleans up the context. Jitter in a live-but-slow connection (data
+	// arriving in small delayed bursts, well under streamStallTimeout) is
+	// handled further up the chain by livePrefetchStreamer, which decodes
+	// off the audio-callback goroutine so this Read never blocks the
+	// speaker's mutex — see its doc comment for the full rationale.
 	var body io.ReadCloser = &stallReader{rc: resp.Body, cancel: cancel, timeout: streamStallTimeout}
 
 	// Wrap in ICY reader if the server provides a metaint interval.
--- a/player/pipeline.go
+++ b/player/pipeline.go
@@ -34,7 +34,10 @@ type trackPipeline struct {
 	// clobbering a newer manual selection.
 	gaplessToken uint64
 
-	live         bool
+	live bool
+
+	// livePrefetch is set when stream is wrapped in a livePrefetchStreamer
+	// (live/non-seekable HTTP sources). close() stops its fill goroutine.
 	livePrefetch *livePrefetchStreamer
 }
 


OtavioPompolini:
Follow-up fix pushed: the initial `livePrefetchStreamer` padded short reads with a hard cut to silence (no fade), which is audible as a click on every partial read — common, not rare, under the exact sustained-slow-connection condition this streamer exists for. Added a short (~1.5ms) linear fade on both sides of every silence boundary. Re-verified against the same live stream with the fade in place.

Issue #348, opened 2026-08-24
Lyrion Music Server (LMS) provider


CaseyRo:
## Problem

cliamp supports seven self-hosted / remote music backends (Navidrome, Plex, Jellyfin, Emby, Audiobookshelf, Qobuz, Tidal) but not Lyrion Music Server (LMS, formerly Logitech Media Server) — the long-running server behind the Squeezebox ecosystem, still actively maintained and very widely deployed on home NAS boxes.

Today an LMS user has no way to reach their library from cliamp, and there's no bridge that closes the gap:

- The LMS **Subsonic plugin** goes the wrong direction — it lets LMS *consume* an external Subsonic library, it doesn't expose the LMS library over the Subsonic API. So the existing Navidrome provider can't be pointed at it.
- The LMS **UPnP bridge** can't adopt cliamp, since cliamp isn't a UPnP renderer (that was raised in #117, which was closed by pointing at the Plex provider).

## Proposal

Add a `lyrion` provider that treats LMS as a library backend, exactly like the other remote providers: browse over its JSON-RPC API, play through cliamp's own engine.

This fits the existing provider seam very cleanly:

| cliamp interface | LMS query |
|---|---|
| `playlist.Provider` | `playlists`, `playlists tracks` |
| `provider.Searcher` | `titles` with `search:` |
| `provider.ArtistBrowser` | `artists`, `albums artist_id:` |
| `provider.AlbumBrowser` | `albums` with `sort:` + offset/count |
| `provider.AlbumTrackLoader` | `titles album_id: sort:albumtrack` |

Playback needs no new machinery — LMS serves the original file at `/music/<track_id>/download`, so tracks are ordinary `Stream: true` HTTP tracks that route through the existing buffered pipeline alongside Subsonic/Jellyfin/Plex. No new dependencies: it's plain JSON over HTTP.

Deliberately **not** proposed: SlimProto / Squeezebox player emulation. That would invert cliamp's architecture (the server would drive playback and own the queue) and is a much bigger project — anyone who wants sync groups and multi-room is better served running `squeezelite`.

Known limitations worth stating up front:

- LMS records play counts against a *player* playing a track. Since cliamp plays the file directly rather than through an LMS player, cliamp listening leaves no history on the server. There's no honest `PlaybackReporter` mapping, so it doesn't implement one.
[...]

PR #349, opened 2026-08-24
feat(provider): add Lyrion Music Server (LMS) provider

CaseyRo:
## Summary

Adds a `lyrion` provider for [Lyrion Music Server](https://lyrion.org/) (LMS, formerly Logitech Media Server) — the server behind the Squeezebox ecosystem. Closes the gap described in #348.

LMS is treated as a **library backend**, exactly like the other remote providers: the library is browsed over the JSON-RPC endpoint (`POST /jsonrpc.js`) and tracks play from the HTTP file endpoint (`/music/<track_id>/download`) through cliamp's own engine. Deliberately *not* SlimProto player emulation — that would invert the architecture and is a much larger project.

Implements `playlist.Provider` plus `Searcher`, `ArtistBrowser`, `AlbumBrowser`, and `AlbumTrackLoader`. No new dependencies — it's plain JSON over HTTP on the standard library.

Three things worth a reviewer's attention:

**Credentials never reach a track.** `Track.Path` is persisted verbatim to resume state and to the play history, and LMS authenticates with the user's actual account password rather than a revocable token (unlike Plex/Navidrome) — so putting credentials in the stream URL would write that password to disk in plaintext. Tracks carry a bare `lyrion://track/<id>`; a registered `SourceResolver` applies credentials at play time, the same shape the Tidal provider uses for its signed URLs. `IsStreamURL` therefore matches the *resolved* URL, not the track path.

**Plugin-contributed content is hidden by default.** An LMS library can mix file-backed tracks with tracks added by server plugins (Spotty and similar). LMS accepts `/music/<id>/download` for those and then never sends a byte — the connection just hangs.

I first returned them flagged-but-visible, reasoning that hiding library content misrepresents the library. Live testing killed that: **all 114 saved playlists** on the test server were Spotty imports, so the provider pane was entirely entries that did nothing. Plugin tracks and plugin-imported playlists are now omitted by default, with `show_unplayable = true` restoring the flagged-but-visible view. Filtering playlists is free — the `playlists` query returns each one's origin URL in the same response, so one added `tags:u` separates `spotify:playlist:…` from `file://…`.

Documented gap: an album made entirely of plugin tracks still appears and opens empty (41 of 1,293 on the test server). The `albums` query carries no origin field, so classifying albums would cost ~1,300 extra requests.

**No `PlaybackReporter`.** LMS records play counts against a *player* playing a track. Since cliamp plays the file directly, there's no honest report to send; synthesising a fake player belongs to the emulation approach, not this one. Documented in `docs/lyrion.md` so it isn't mistaken for a bug.

One subtlety that cost me a bug and is now pinned by a test: the `albums` query uses **uppercase `S`** for `artist_id`, while the `titles` query uses **lowercase `s`** for the same field (lowercase `s` in `albums` is a sort textkey). Since LMS silently omits any field whose tag letter you didn't ask for, getting that backwards returns every album with a blank artist and no error anywhere.

## Screenshots / video

No UI changes — the provider appears in the existing provider panel and browser alongside the others.

## How to test

1. Point it at an LMS instance, either via `cliamp setup` (there's a new Lyrion entry) or by hand:

   ```toml
   [lyrion]
   url = "http://nas.local:9000"
   # user/password only if the server has password protection
   # show_unplayable = true   # include plugin-contributed tracks and playlists
   ```

   `LYRION_URL` / `LYRION_USER` / `LYRION_PASS` work as a fallback when no block is present.

2. Launch `cliamp` and confirm Lyrion appears in the provider panel; browse playlists, artists, and albums, search, and play a track.

3. There's an env-guarded live test suite following the `netease` precedent:

   ```sh
   CLIAMP_LYRION_LIVE_URL=http://nas.local:9000 go test ./external/lyrion/ -run Live -v
   ```

   It asserts metadata actually comes back populated (which is what catches a wrong tag set) and streams real audio bytes. Verified against LMS 9.2.0 with 16,663 tracks / 4,225 artists.

## Checklist

- [x] `make check` passes
- [x] `docs/` and `site/index.html` updated for user-facing changes

New `docs/lyrion.md`; Lyrion added to the provider lists in `README.md`, `docs/configuration.md`, and `site/index.html`.

---

Happy to adjust naming, the `Unplayable` heuristic, or anything else to taste — and to drop the live test file if you'd rather not carry one.

## Summary by CodeRabbit

* **New Features**
  * Added Lyrion Music Server as a supported music provider.
  * Browse playlists, artists, albums, and tracks, with track search.
  * Stream compatible tracks with optional authentication.
  * Configure Lyrion through the setup wizard, configuration files, or environment variables.
  * Optionally display unplayable plugin-provided tracks.

* **Documentation**
  * Added setup, configuration, limitations, security, and troubleshooting guidance.
  * Updated supported-provider listings across the README and website.

* **Tests**
  * Added coverage for configuration, setup, browsing, searching, authentication, and streaming.


CaseyRo:
Thanks — the credentials-in-URL finding was a real bug and a good catch. Pushed 08009de addressing it and most of the rest.

**Credentials in `StreamURL` — fixed, and it was worse than the review suggested.** `Track.Path` isn't just passed to the player: it's persisted verbatim to resume state (`internal/resume`) and to the play history (`history/history.go` writes `path = %q`). And unlike Plex or Navidrome, which embed a *revocable token*, LMS authenticates with the user's actual account password — so this was writing that password to two files in plaintext on disk.

The suggested fix (authenticate via "the client's existing authenticated transport") isn't quite available, since the player fetches the URL with its own HTTP client rather than the provider's — that's why the userinfo was there originally. So I used `RegisterSourceResolver`, which exists for exactly this and is how the Tidal provider handles its short-lived signed URLs:

- `Track.Path` is now `lyrion://track/<id>` — no credentials, no host.
- A registered resolver expands it to the authenticated URL when playback starts, so credentials live only in memory.
- `IsStreamURL` consequently matches the *resolved* URL, not the track path.

Verified on disk: `resume.json` now records `{"path":"lyrion://track/19070"}`, and playback still works end to end against a live LMS 9.2.0 instance. Regression tests added in both the provider and wiring packages.

**Also fixed:**
- Setup test now asserts the `user`/`password` fields are *present* and optional, rather than only that they aren't required — a dropped field would previously have passed silently.
- Live test uses the package `httpClient` so it inherits the configured timeout.
- Docs: softened the absolute "never delivers any audio" to "may never deliver any audio, leaving playback to hang" (I only observed this on Spotty-contributed tracks), and took the grammar fix.

**Declined:** wrapping the `Ping()` error in the setup validator. Every other provider in `cmd/setup.go` returns the bare `Ping()` error (`navidrome:118`, `plex:166`, `jellyfin:201`), and the Lyrion error already carries `lyrion: <url>: …` context from the request layer, so wrapping would add a second prefix and make this the only inconsistent entry.

Separately, while testing I found and fixed an unrelated gap in my own change: `--provider lyrion` was rejected by the allowlist in `overridesFromFlags` even though the docs listed it as valid, so `default_provider = "lyrion"` failed. Fixed with a regression test.

CaseyRo:
Addressed the remaining three findings in b2d06ab.

- **Table-driven track mapping** and **table-driven `show_unplayable` config test** — both converted, matching the repo's stated preference in `CLAUDE.md`. The config test gained a third case for an explicit `show_unplayable = false`, which wasn't covered before.
- **Live album probe pagination** — this one was a genuine flakiness risk in a test I'd written, not just style. Albums made entirely of plugin tracks filter to empty and cluster together in sort order, so a server with enough of them at the top of its catalogue would have failed the test spuriously. It now pages through until it finds a playable album or exhausts the catalogue.

`make check` is green across 47 packages, and the live suite still passes against LMS 9.2.0.

bjarneo:
Can you fix the merge conflicts?

CaseyRo:
Done — merged `main` in at 16890a4, branch is mergeable again.

All eight conflicts were the same shape: the Mixcloud provider (#344) registers at the same points Lyrion does, so both entries now sit side by side — meta keys in `provider/types.go`, imports in `main.go` and `cmd/setup.go`, the `--provider` usage string and validation in `commands.go`, `config/config.go`, and the provider lists in `README.md`, `docs/configuration.md`, and `site/index.html`. Where `main` had also reworded the surrounding prose I kept your wording and slotted Lyrion into it.

One thing worth flagging on its own: **`main` currently doesn't compile.** #328 (9c827dc) added an early exit to `applyResume`, which returns `tea.Cmd`:

```
ui/model/playback.go:610:3: not enough return values
	have ()
	want (tea.Cmd)
```

I reproduced it on a clean `origin/main` worktree, so it isn't something the merge introduced. Since merging `main` inherits the break, I fixed the bare `return` → `return nil` as a separate commit (01c09e6) so this branch builds — happy to drop it if you'd rather land that on `main` directly.

`make check` is green across all packages.

PR #350, opened 2026-08-25
fix(ui): load direct stream URLs from the u overlay

joe25211:
## Summary

Pressing `u` and entering a direct audio or stream URL fails with `resolving feed <url>: parsing feed: XML syntax error`. The same URL works as a command-line argument, and the `url.load` IPC operation loads it into the very same running instance.

`docs/streaming.md` already documents the intended behaviour:

> Press `u` while playing to load a new stream or playlist URL without restarting. Supports the same URL types as CLI arguments: direct audio URLs, M3U/PLS playlists, RSS podcast feeds, and yt-dlp compatible links.

**Cause.** `handleURLInputKey` (`ui/model/keys.go`) hands its input to `resolveRemoteCmd`, which calls `resolve.Remote`. `Remote`'s `default:` arm assumes the URL was already classified as a feed:

```go
default:
	// URL was classified as a feed by content-type sniffing.
	t, err := resolveFeed(u)
```

I grepped every `resolve.Remote` call site in the tree. The overlay is the only one that doesn't establish that precondition:

| call site | input |
|---|---|
| `ui/model/init.go:214` | `m.pendingURLs`, produced by `resolve.Args` |
| `ui/model/commands.go:342` `resolveWrapperURLs` | guarded on `IsPLS(t.Path) \|\| IsM3U(t.Path)` |
| `ui/model/commands.go:271` `fetchNetSearchCmd` | an `ytsearch10:`/`scsearch10:` prefix, matched by `IsYTDL` via `IsYTSearch` |
| `ui/model/plugin_queue.go:92` | `Args`' `Pending` |
| **`ui/model/keys.go:1575` (the `u` overlay)** | **raw text typed by the user** |

Both handlers for the `url.load` operation already call `resolve.URL` instead — `daemon.handleURL` (`daemon.go:263`) headless, and `Model.handleIPCURL` (`ui/model/ipc_extended.go:35`) into a running TUI. So a stream URL sent as `url.load` already worked while `u`, in the same process, did not.

**Fix.** `resolve.URL` already does exactly what the overlay needs — it classifies through `Args` before falling back to `Remote`, and `resolve/args_test.go:165` (`TestURLResolvesRawStream`) already pins it handling a raw stream. The overlay now calls it.

`resolveRemoteCmd` is deliberately left alone. Fixing `Remote`'s `default:` arm instead would make the startup path in `init.go` sniff every URL a second time, since `Args` already sniffed them to decide they were feeds.

`Remote`'s doc comment now states that precondition, so the next caller doesn't have to infer it from a comment inside the switch.

No `docs/` or `site/index.html` change: this makes the code match what `docs/streaming.md` already says, and `site/index.html` doesn't cover the overlay.

## Screenshots / video

Not included — the change has no visual component beyond the overlay no longer erroring, and the repro below is runnable. Happy to add a recording if you'd prefer one.

## How to test

1. Start cliamp, press `u`, and paste any direct audio URL (an `.mp3` link, or an extensionless Icecast/Shoutcast stream). Before this change it fails with `parsing feed: … XML syntax error`; now it loads and plays.
2. Press `u` and paste an RSS podcast feed or an M3U/PLS URL — unchanged, still resolves.
3. Pass the same stream URL as a CLI argument, or send it to a running instance as the `url.load` IPC operation — both already worked. That asymmetry was the bug.
4. `go test ./ui/model/ ./resolve/`

Automated coverage:

- `TestURLOverlayLoadsRawStream` (`ui/model/commands_test.go`) drives `handleURLInputKey` with Enter rather than calling the command directly, so reverting the call site fails it.
- `TestRemoteRequiresClassifiedURLs` (`resolve/resolve_test.go`) pins the precondition the new doc comment describes.

Five mutations were applied to the new assertions — reverting the call site, pointing the new command back at `resolve.Remote`, dropping the source URL from `feedsLoadedMsg` (which `initYTDLBatch` reads), dropping `autoPlay`, and relaxing `Remote`'s `default:` arm. All five failed the suite.

## Checklist

- [x] `make check` passes — also `make ci` (staticcheck 0, govulncheck 0 affecting) and `go test -race -count=1 ./...` across all 46 packages. Windows was checked by cross-compiling the touched packages (`GOOS=windows go build ./resolve/ ./ui/model/`); the full-tree Windows build needs CI's MinGW/cgo codec toolchain, which can't be reproduced from Linux — `GOOS=windows go build ./...` fails identically on unmodified `main`, so it is not a signal either way.
- [x] `docs/` and `site/index.html` updated for user-facing changes — no change needed; see Summary

## Summary by CodeRabbit

* **New Features**
  * Added support for loading raw stream URLs directly from the URL input overlay.
  * Raw stream URLs now resolve asynchronously and can start playback automatically.

* **Bug Fixes**
  * Improved URL handling to correctly distinguish unclassified URLs from remote playlists.

* **Tests**
  * Added coverage for raw stream URL loading and URL resolution behavior.


bjarneo:
Can you fix the merge conflict?

PR #368, opened 2026-08-26
fix(spotify): use an extended-quota client ID for the Web API

### LINKED ISSUE #214: [Spotify]  Followed playlists unplayable for users with post-Nov-2024 Spotify Developer apps — adopt the ncspot/spotify-player fallback `client_id` pattern?
## Summary

If you create a fresh Spotify Developer app today and plug it into `[spotify] client_id = ...`, you can't list or play tracks for any followed playlist. PR #94 papered over this by filtering followed playlists out, but the real cause is a Spotify policy change, not anything cliamp can fix at the per-playlist level.

This isn't really a bug report — more of a request for direction, since I'm happy to do the implementation. `ncspot` and `spotify-player` both sidestep this by shipping a fallback `client_id` baked into the binary, and I wanted to check if you'd be open to that pattern in cliamp before writing any code.

## Root cause

Spotify's [Nov 27, 2024 Web API changes](https://developer.spotify.com/blog/2024-11-27-changes-to-the-web-api) clamped down on newly-created developer apps that don't have "extended quota mode." The blog post calls out algorithmic and Spotify-owned editorial playlists, but in practice the lockdown hits most followed playlists for any dev-mode app. The `playlist-read-private` and `playlist-read-collaborative` scopes don't help — this is enforced at the app-quota layer, not OAuth.

Apps that already had extended quota mode before the cutoff are fine. Getting it now means a written business-case submission, which isn't realistic for end users.

## How other Spotify TUIs handle it

I checked the other terminal Spotify clients:

| Tool | Status | Web API client_id strategy | Affected by Nov 2024 restrictions? |
|------|--------|----------------------------|---|
| spotify-tui | Dormant since 2021 | User-supplied | Yes (same as cliamp) |
| ncspot | Active | Hardcoded `NCSPOT_CLIENT_ID = "d420a117a32841c2b3474932e49fb54b"` (ncspot's own registered app) | No |
| spotify-player | Active (~6.6k⭐) | **Borrows the same `NCSPOT_CLIENT_ID`** as default (`spotify_player/src/auth.rs:7`, `config/mod.rs:297`) | No |
| spogo | Active | None — imported browser cookies + Spotify's internal Pathfinder GraphQL | No (totally different architecture) |

`ncspot` and `spotify-player` ship the same hardcoded `client_id`, registered to the ncspot project — predates the Nov 2024 cutoff and/or has extended quota. Users get full access with no developer-app setup at all. `spotify-player` makes it overridable: if you set your own `client_id`, yours wins; otherwise it falls back to the ncspot ID.

`spogo` is in a different category (cookie-based, internal GraphQL). Mentioning it for completeness, but I noticed cliamp dropped the equivalent spclient fallback in v1.43.0, so I'm guessing that route is already off the table.

## Proposed direction

If the ncspot/spotify-player pattern is something you'd accept:

1. Bake a fallback `client_id` into `external/spotify/`. Two flavors:
   - Use the ncspot/spotify-player ID (`d420a117…`). Quick to land but means cliamp shares another project's rate-limit pool.
   - Register a new app for cliamp specifically and apply for extended quota. Cleaner long-term but it's on you (or whoever owns the project handle) to register and maintain it.
2. Prefer the user's `[spotify] client_id` if set; otherwise use the fallback.
3. Drop the `playlistAccessible()` filter in `provider.go` — the "Followed playlists" section is already wired up in `Playlists()`, it just has nothing to show today.
4. Update the docs so `client_id` is genuinely optional.

If you'd rather not go this route (don't want to ride on someone else's app, don't want to register a new one, etc.), that's totally fair — happy to close this with a note pointing people at ncspot/spotify-player/spogo, and leave the filter where it is.

## Question

Would you accept a PR that adds a fallback `client_id` (either borrowed or cliamp-registered) and drops the followed-playlist filter? Or do you have a different direction in mind? Happy to do the implementation once you let me know.

abguymon:
## Problem

cliamp uses the librespot keymaster `client_id` (`65b70807…`) for **both** playback and the Web API. `docs/spotify.md` describes this as the identity "librespot and spotify-player use" — that's true for playback, but not for the Web API, and the difference matters:

- **Rate limits.** Spotify applies quota per `client_id` globally. Keymaster is shared by every librespot-based player in existence, so its Web API pool is permanently hot and `429 Too Many Requests` on search and playlist loading is routine rather than occasional.
- **Endpoint access.** Anyone who follows the current docs and registers their own app lands in Development Mode, which cannot read playlists the user follows but does not own. Those requests fail with `403 Forbidden`.

Since dd04d28 removed the `playlistAccessible()` ownership filter, followed playlists are listed in the library — so the 403 is now reachable straight from the UI. That commit also removed the branch that turned it into a readable message, so it surfaces as a bare `spotify: list tracks: http status 403 Forbidden`.

This is the underlying issue in #214.

## What other clients do

ncspot keeps two distinct identities, and spotify-player defaults to the same pair:

| Purpose | Client ID | Source |
| --- | --- | --- |
| Playback | keymaster `65b70807…` | [`ncspot/src/spotify.rs:152`](https://github.com/hrkfdn/ncspot/blob/main/src/spotify.rs#L152) |
| Web API | ncspot `d420a117…` | [`ncspot/src/spotify_api.rs:49`](https://github.com/hrkfdn/ncspot/blob/main/src/spotify_api.rs#L49) |

ncspot's `client_id` is registered in Extended Quota Mode. It predates the [November 2024 changes](https://developer.spotify.com/blog/2024-11-27-changes-to-the-web-api) and is explicitly exempt from the [February 2026 ones](https://developer.spotify.com/documentation/web-api/tutorials/february-2026-migration-guide) — *"Apps in extended quota mode are not affected by any of the changes described in this guide."* spotify-player's README recommends against overriding it: *"You almost certainly should not configure your own `client_id`."*

## Changes

- Split the constant into `PlaybackClientID` (keymaster, still mints the `streaming` grant) and `DefaultClientID` (ncspot's, used for the Web API when the user hasn't set their own).
- `interactiveOAuthFlows` now keys the single-flow path off `PlaybackClientID`. The default path becomes two flows in one browser tab — web api, then playback — which is what a custom `client_id` already did.
- `isExtendedQuotaClient` replaces the direct `DefaultClientID` comparison guarding the one-request search path, so both built-in identities keep it.
- Docs, `config.toml.example`, and the setup wizard recommended registering a personal app. They now recommend the built-in identity and state the Development Mode restrictions plainly; the wizard's picker defaults to it.

## Migration

Existing users sign in once more. The stored refresh token was issued to keymaster and won't refresh against the new Web API client; the existing `invalid_grant` path clears credentials and prompts re-auth.

## Testing

`go build ./...`, `go vet ./...`, `gofmt -l .` clean; `go test ./...` passes (50 packages).

Updated `TestInteractiveOAuthFlows` for the new split and added `TestBuiltInClientIDsAreDistinct` to keep the two identities from being collapsed again. `TestPasteIntoActiveField` needed a `KeyDown` because the wizard's recommended option moved to the top.

Verified against a real account: a followed playlist that returned `403 Forbidden` under a self-registered Development Mode app, then `429` under keymaster, loads normally on the Extended Quota Mode client.

Closes #214.

## Summary by CodeRabbit

- **New Features**
  - Spotify setup now recommends built-in credentials, with separate authorization for Web API access and playback.
  - Improved search paging behavior for supported built-in credentials.

- **Bug Fixes**
  - Improved handling of Development Mode restrictions and shared-quota rate limits.
  - Custom Web API credentials no longer request unnecessary playback permissions.

- **Documentation**
  - Updated setup, configuration, and troubleshooting guidance with credential recommendations, playlist limitations, authorization steps, and possible errors.
  - Clarified that custom Spotify apps are an advanced option.


k7cfo:
Field notes from a machine running Omarchy's packaged cliamp **v1.63.2** with a personal Development Mode `client_id` (playlists/playback already worked; search did not).

**What we hit**

- Packaged `SearchTracks` asks `/v1/search` for up to 50 results. Development Mode returns `400 Invalid limit` at >10. v1.63.2 then rewrites that as “client_id is too new / search blocked” and tells you to drop `client_id`.
- Dropping `client_id` and using the **keymaster** identity (`65b70807…`) was worse here: `/v1/me` came back **429** with retries, so the library never loaded.
- Paging `/v1/search` at **limit 10 + offset**, while keeping the personal `client_id`, made search work (`johnny cash` → titled tracks) and left playlists intact (`Your Music` + the rest). Same result after a restart.

That paging path is already on `main` via #315 and #337. Omarchy is just still on 1.63.2, so this is a packaging lag, not a missing `main` fix. I would **not** open another paging PR against `main`.

**How this PR (#368) differs — and why it is still useful**

This change is not the same as paging. It splits:

| Purpose | Identity |
| --- | --- |
| Web API | ncspot Extended Quota `d420a117…` when `client_id` is unset |
| Playback | keymaster `65b70807…` (streaming grant) |

That matches what we saw: keymaster as a **Web API** client is a global 429 pool. Paging does not fix that. For users who follow the old docs and omit `client_id`, this PR is the right fix.

**What it does not replace**

- A **personal Development Mode** `client_id` still needs #315/#337 paging (already on `main`). Search worked for us that way without switching to ncspot's ID.
- Followed-but-not-owned playlists `403` is a Development Mode restriction paging cannot fix. If that is the goal, the built-in Extended Quota identity (this PR) is the actual lever — with the usual caveat that you are joining ncspot/spotify-player's shared quota.

**Review nits**

- Treating **keymaster** as `isExtendedQuotaClient` for the one-shot search path is a bit optimistic. On this box keymaster was 429-limited on `/v1/me` long before search. Falling back to `searchPaged` only on `Invalid limit` will not help a 429. Retry/backoff already exists; just do not assume keymaster search is “the fast path.”
- Shipping ncspot's client ID is the same bet ncspot/spotify-player already make. Worth stating in the PR that a Spotify clamp/revocation of that ID would land on every cliamp user who took the new default, and that existing keymaster refresh tokens need a one-time re-auth (you already documented this).
- Two OAuth steps for the new default is the correct model; the setup-wizard copy change to “built-in recommended” is consistent with that.

**Suggestion:** land #368 for the default/no-`client_id` path; keep personal apps on paging from `main`; do not revert people with a working private `client_id` onto the shared Web API pool unless they need followed-playlist access.

Happy to re-test a build of this branch against the same account if useful.

abguymon:
Rebased on main and was conflicting in docs/spotify.md (with #461 and #467). Kept main's wording on both.

@k7cfo both nits were fair, thanks.

The first one was really a naming bug. isExtendedQuotaClient sounded like it was making a claim about quota, but the only thing that branch decides is whether /v1/search will take more than 10 results in one request — a Development Mode question. It's now allowsFullSearchPage, and the comment says outright that keymaster is a bad Web API identity that'll hit 429 long before it ever sees Invalid limit. I left the fallback keyed to Invalid limit deliberately: paging a 429 would mean more requests, and webAPI already backs off and honors Retry-After. Behavior didn't change.

Second one is now written down in the docs and config.toml.example — neither built-in client ID is ours, so if Spotify clamps or revokes one, everyone on the default is affected at the same time. Same bet ncspot and spotify-player are making, but better said out loud than implied.

Your split between this and #315/#337 is the right framing: paging fixes the Development Mode 400, this fixes the keymaster 429 and the followed-playlist 403. A personal client ID still needs the paging.

One thing to ignore: CodeRabbit's "Linked Issues" check says the followed-playlist filter wasn't removed. It was, in dd04d28 — that's what made the 403 reachable from the UI to begin with.

Build, vet, gofmt, and tests all clean.

PR #374, opened 2026-08-27
fix(nix): downgrade Go version to one that is in nixpkgs

richardgill:
`nix run github:bjarneo/cliamp` currently fails because the project requires a newer Go patch release than the pinned nixpkgs provides:

```text
go: go.mod requires go >= 1.26.6 (running go 1.26.5; GOTOOLCHAIN=local)
```

This lowers the minimum Go version to 1.26.5, matching nixpkgs unstable.

Verified with:

```sh
nix build . --no-link
```

## Summary by CodeRabbit

* **Chores**
  * Updated the required Go toolchain version for improved compatibility with Go 1.26.5.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,16 +2,16 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
-        "lastModified": 1787816113,
-        "narHash": "sha256-jQieHgYZkOGlNFt1hadB47o8pj0BYCP7nDtZw8pJw88=",
+        "lastModified": 1787135253,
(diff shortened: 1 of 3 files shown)


gjermundgaraba (changes requested):
I dont believe this is such a good idea. as code rabbit says, this would a security misconfiguration. if you have a solution that instead upgrades the nixpkgs go version, that would be better.

gjermundgaraba:
closing this for now, feel free to re-open if you have a non-downgrading solution :)

PR #379, opened 2026-08-27
Expose live radio track title (broadcast_title) over IPC status

wsturgiss:
Player.StreamTitle() already unifies ICY inline metadata and the radiometa station-API fallback, and cliamp's own TUI displays it live, but neither status-response builder (daemon.go's headless path, or ui/model/update.go's interactive-TUI path) put it on the wire. Add it to ipc.TrackInfo as broadcast_title, set from Player.StreamTitle() in both builders when the track is a stream.

Verified against live playback (Radio Motown, real ICY metadata confirmed via a standalone probe of the stream) using a locally built binary as both client and server.

## How to test

1.

## Checklist

- [ ] `make check` passes

### DIFF
--- a/daemon.go
+++ b/daemon.go
@@ -956,6 +956,9 @@ func (d *daemon) statusResponse() ipc.Response {
 	}
 	if cur, _ := d.playlist.Current(); cur.Path != "" {
 		info := trackInfo(cur, d.playlist.Index(), d.playlist.QueuePosition(d.playlist.Index()))
+		if info.Stream {
+			info.BroadcastTitle = d.player.StreamTitle()
+		}
 		resp.Track = &info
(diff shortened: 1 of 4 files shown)


wsturgiss:
Closing — this is superseded by #319, which already merged into `main` and ships the same feature (live radio now-playing title over IPC status) via `TrackInfo.StreamTitle`/`Artist`/`Station` and the shared `applyStreamTitle()` helper in `daemon_v2.go`, as part of the v2 IPC refactor. That refactor also removed `ipc.StatusRequestMsg` and the old `statusResponse()`/inline status-building code this PR patches, so rebasing would just reintroduce dead code rather than add anything new.

PR #389, opened 2026-08-30
feat(ipc): expose the active screen and its keymap

seantimm:
## Summary

cliamp already knows which screen owns keyboard input and builds an exact per-screen keymap for the `Ctrl+K` overlay — filtered to the commands valid on that screen, minus anything currently disabled, plus Lua plugin bindings. Nothing outside the process could read either. This exposes both:

- **`screen` on the `state.get` snapshot** (and `status --json`): `{"id":"file_browser","label":"Files"}`. Every `commandMode` gets a stable id; a drift-guard test fails if a new mode is added without one.
- **`keymap.get`**, a read method like `state.get`: the overlay's rows as `{keys, label, action, section}`. `keys` carries the Bubbletea key names a client can *send* (`"space"`, `"ctrl+k"`), separate from the human label (`"Left Right"`). `section` is `current` / `main` / `plugins`, mirroring the overlay's dividers.
- **`cliamp keymap [--json]`**, a thin client mirroring `status`.

The overlay and the export share one walker (`walkKeymap`), so they cannot drift — a test asserts their row counts match, and the existing keymap tests pass unchanged against the refactor. `--daemon` answers `keymap.get` with `unavailable` and a detail string rather than an empty list a client might read as "no keys".

Motivation: a desktop-level "show me this app's keybindings" menu that asks the focused app what applies right now. Neovim and tmux can answer that; cliamp's screens rebind the same keys, so without this an external tool can only show the main screen's list, which is wrong whenever a picker or browser is open.

### Prior art

Each checked against the installed version, not from memory:

- **mpv** — `input-bindings` over its JSON IPC returns every active binding as `{key, cmd, section, …}`; the `section` field is the same per-screen distinction.
- **Neovim** — `nvim_get_keymap(mode)` + `mode()` over RPC.
- **tmux** — `list-keys -T <table>` with `#{pane_in_mode}` telling a client which table applies.
- **GTK/libadwaita** — shortcuts shipped as an `AdwShortcutsDialog` resource.

## Screenshots / video

No UI change. Live output from a dev build, main screen → `o` → file browser:

```
$ cliamp status --json | jq -c .screen
{"id":"playlist","label":"Playlist"}
$ cliamp keymap
[main]
  Space              Play / Pause
  ...
# press o
$ cliamp status --json | jq -c .screen
{"id":"file_browser","label":"Files"}
$ cliamp keymap
[current]
  Esc                Back
  Enter              Select
  R                  Replace queue
[main]
  Space              Play / Pause
```

## How to test

1. `make check`
2. Run the TUI, then from another shell: `cliamp status --json | jq .screen` and `cliamp keymap`
3. Press `o` (file browser), `p`, `A`, `y`, `t` in the TUI and re-run both — `screen.id` changes and a `[current]` section appears first
4. `cliamp --daemon` then `cliamp keymap` → `unavailable` with a detail message

## Checklist

- [x] `make check` passes
- [x] `docs/` and `site/index.html` updated for user-facing changes (`remote-control.md`, `cli.md`, `keybindings.md`, site Remote Control blurb)

Note: `ipc/v2.go` shows a large diff that is one inserted field plus gofmt realigning the struct's type column (`git diff -w` shows 1 insertion).

---
*Opened by Claude Code on @seantimm's behalf.*

## Summary by CodeRabbit

- **New Features**
  - Added `cliamp keymap` to display commands available on the active screen.
  - Added `--json` output for machine-readable key bindings.
  - Remote-control snapshots now include the active screen.
  - Added the `keymap.get` remote-control endpoint.
  - Key bindings now follow the current playlist, file browser, or queue view.

- **Documentation**
  - Updated CLI, keybindings, remote-control, and feature documentation.
  - Documented that keymaps are unavailable when running in headless daemon mode.


bjarneo:
What is the use case?

seantimm:
I'm setting up a central keybinding system for Omarchy (https://github.com/omacom/omarchy/pull/9093) to show available bindings based on your screen context.  Rather than hardcode default keybindings, I set it up to be dynamic based on the available mappings, so when cliamp is the active window, it would show keybindings for cliamp.  I can only show the base bindings right now because I don't know the active context for the app.

PR #394, opened 2026-08-30
fix(player): follow the audible position so waveform visualizers reach the tick rate

### LINKED ISSUE #392: Waveform visualizers (Wave/Scope/Heartbeat) repaint at ~8.5 FPS despite TickWave = 16 ms
### Summary

The waveform modes ask for `TickWave = TickAnim = 16 * time.Millisecond` (~60 FPS) in `ui/tick.go`, but on screen they update roughly 8-9 times per second — noticeably less smooth than `Bars`, which asks for the slower `TickFast` (50 ms) and hits it exactly.

### Measurements

cliamp 1.63.2 (Arch, `cliamp-bin`), playing a local 320 kbps MP3, foot terminal on Hyprland/Wayland, window 1416x1243 px (full layout, five-row visualizer), focused. I counted the actual `write()` calls cliamp makes to its pty over 6 seconds with `strace -f -e trace=write` and filtered on fd 1 (frames of ~766 bytes on average, so these are real repaints, not stray control sequences):

| Visualizer | writes in 6 s | effective FPS | requested cadence |
|---|---|---|---|
| Wave | 51 | **8.5** | `TickWave` = 16 ms (~60 FPS) |
| Heartbeat | 52 | 8.7 | `TickWave` = 16 ms (~60 FPS) |
| Bars | 120 | **20.0** | `TickFast` = 50 ms (20 FPS) |

Switching between the three was done with `cliamp vis <mode>` on the running instance, everything else identical.

Things I ruled out:

- **Volume / `vis_volume_linked`**: measured at -6 dB and at 0 dB, 8.3 and 8.5 FPS — no difference, so it isn't a flat waveform being diffed away because of amplitude.
- **`low_power`**: not set (`config.toml` has only theme/visualizer/eq/repeat).
- **Window size**: the full layout was active, the five-row visualizer was rendering.
- **Mode switch**: Wave and Heartbeat measure the same, so this isn't specific to one renderer.

### Guess at the cause (unverified)

Wave, Scope and Heartbeat are registered with `newFastRenderOnlyDriver(spectrumAnalysisSpec(0), TickWave, ...)` and read `v.waveBuf`, which is refreshed from raw samples. If that buffer is only refreshed at around 8-9 Hz, the frames in between render to an identical string and Bubbletea's diffing renderer writes nothing — the tick would be running at 60 Hz while the picture is not. `Bars` goes through the FFT path, which appears to produce new data every tick. I could not find the code that sets the sample-refresh rate, so this is a hypothesis, not a diagnosis.

### Why it matters

`TickWave` exists specifically to make waveform modes smooth, and it's the one place where the cadence constant and the result disagree. Right now the way to get a smoother visualizer is to pick a mode that asks for a *slower* tick.

Happy to run more measurements on this machine if that helps — an instrumented build counting `waveBuf` refreshes would settle the cause quickly.

rylos:
Fixes #392.

### The problem

`Wave`, `Scope` and `Heartbeat` are registered with `TickWave` (16 ms, ~60 FPS), but on screen they repaint about 7-8 times per second, while `Bars` — which asks for the *slower* `TickFast` — hits its 20 FPS exactly.

The cause is in `player/tap.go`. `WaveformSamplesInto` anchored the read position to the **last `Stream` call** and clamped the elapsed time to that call's frame count:

```go
frames := t.writeFrames.Load()
elapsed = max(0, min(elapsed, frames))
end := t.written.Load() - frames + elapsed
```

The backend doesn't pull samples at a steady rate: it refills its buffer in bursts and then goes quiet for as long as that buffer takes to play, so the final `Stream` call of a refill can be a short tail. The window walked through those few frames and then froze until the next burst — tying the effective refresh rate to the **buffer size** rather than to the tick rate.

That is measurable: with `--buffer-ms 50` the same build reaches 19 FPS, with the default 250 ms it drops to 7-8.

### The change

Anchor the position to the backend buffer instead. At the moment of a write, what is audible is `speakerFrames` behind what has been written, and it advances with elapsed wall-clock time:

```go
end := written - int64(t.speakerFrames) + elapsed
end = max(0, min(end, written))
```

`speakerFrames` is the speaker buffer in frames, which `Player.New` already computes for `speaker.Init`; it is passed to `newTap` alongside the ring size. The clamp to `written` means a stalled backend stops the window instead of walking it into stale ring-buffer data, and `speakerFrames <= 0` falls back to the newest frame, so the zero value stays safe.

### Measurements

Arch Linux, foot on Hyprland/Wayland, 320 kbps MP3, `Wave` mode, counting `write()` calls to the pty with `strace -f -e trace=write` over 5 s:

| Buffer | before | after |
|---|---|---|
| 250 ms (default) | 8.2 FPS | **19.4 FPS** |
| 50 ms | 19.0 FPS | 19.0 FPS |

19-20 FPS is the ceiling set by `tea.WithFPS(defaultUIFPS)` in `main.go`, so the waveform modes now reach the renderer's cap instead of the backend's refill rate. Raising that cap to let `TickWave` actually deliver ~60 FPS is a separate decision and is not part of this PR.

### Tests

`go test ./...` passes. Two tests added to `player/tap_test.go`:

- `TestTapWaveformAdvancesPastLastWriteSize` — a long refill followed by a short tail, exactly the shape that caused the freeze; the two reads returned identical frames before this change.
- `TestTapWaveformDoesNotReadPastWritten` — a 30 s stall must leave the window at the newest written frame.

The existing `TestTapWaveformSamplesAdvanceBetweenWrites` keeps its expectations unchanged, with the backend buffer passed explicitly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved waveform visualization timing so it advances smoothly between audio buffer refills.
  * Prevented waveform displays from stalling when audio writes are shorter than the playback buffer.
  * Ensured visualizations remain synchronized with the audible playback position.
  * Prevented waveform displays from reading beyond the audio data currently available.
  * Improved waveform stability when playback pauses or reaches the latest available audio.

### DIFF
--- a/player/player.go
+++ b/player/player.go
@@ -53,6 +53,9 @@ type Player struct {
 	resampleQuality int
 	bitDepth        int // 16 or 32
 	tapBufferFrames int
+	// speakerBufferFrames is the audio backend's buffer, in frames: the latency
+	// between a frame reaching the tap and the same frame being heard.
+	speakerBufferFrames int
 
 	gaplessAdvance atomic.Bool   // set when gapless transition fires
 	seekGen        atomic.Int64  // generation counter for yt-dlp seeks; incremented to cancel stale seeks
@@ -91,10 +94,11 @@ func New(q Quality) (*Player, error) {
 		bitDepth = 16
 	}
 	p := &Player{
-		sr:              sr,
-		resampleQuality: q.ResampleQuality,
-		bitDepth:        bitDepth,
-		tapBufferFrames: max(4096, sr.N(time.Duration(q.BufferMs)*time.Millisecond)),
+		sr:                  sr,
+		resampleQuality:     q.ResampleQuality,
+		bitDepth:            bitDepth,
+		tapBufferFrames:     max(4096, sr.N(time.Duration(q.BufferMs)*time.Millisecond)),
+		speakerBufferFrames: sr.N(time.Duration(q.BufferMs) * time.Millisecond),
 	}
 	p.volMin.Store(math.Float64bits(-50))
 	p.speed.Store(math.Float64bits(1.0))
@@ -302,7 +306,7 @@ func (p *Player) playPipelineForGeneration(tp *trackPipeline, generation uint64)
 			s = newBiquad(s, eqFreqs[i], 1.4, &p.eqBands[i], float64(p.sr))
 		}
 
-		p.tap = newTap(s, p.tapBufferFrames, int(p.sr))
+		p.tap = newTap(s, p.tapBufferFrames, int(p.sr), p.speakerBufferFrames)
 		s = &volumeStreamer{s: p.tap, vol: &p.volume, mono: &p.mono, cachedDB: math.NaN()}
 		p.ctrl = &beep.Ctrl{Streamer: s}
 		p.started = true
--- a/player/player.go
+++ b/player/player.go
@@ -79,6 +79,20 @@ type Player struct {
 // function, the poll interval, and ok=false when the URL is not recognized.
 type StreamMetadataResolver func(streamURL string) (fetch func(ctx context.Context) (string, error), interval time.Duration, ok bool)
 
+// maxAnalysisWindow is the largest window the visualizer asks for in one read
+// (the classic peak meter's 4096-sample FFT). The package cannot import ui, so
+// the value is duplicated here; it only needs to be an upper bound.
+const maxAnalysisWindow = 4096
+
+// tapRingFrames sizes the tap's ring buffer. The waveform read position sits
+// speakerFrames behind the newest frame — that is where audio is actually
+// audible — so the ring has to hold that lag *plus* a full analysis window on
+// top of it. Sized to the backend buffer alone, the oldest end of the window
+// would read frames that have already been overwritten.
+func tapRingFrames(speakerFrames int) int {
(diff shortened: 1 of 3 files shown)


rylos:
Good catch from the bot — that one is real, and fixed in 6f5c21e.

Moving the read position back by `speakerFrames` means the ring has to hold that lag **plus** a full analysis window behind it. With `max(4096, speakerFrames)` the ring was exactly the lag at the default settings (250 ms at 44.1 kHz = 11025 frames), so the oldest end of every window read frames that had just been overwritten.

Added `tapRingFrames`, which reserves the largest window a visualizer asks for in one read — the classic peak meter's 4096-sample FFT. `player` cannot import `ui`, so that bound is duplicated with a comment saying why; if you would rather export it from `ui` and have the player depend on it, or park it in a shared package, I am happy to change it.

Tests added: the sizing for a 250 ms buffer, a 50 ms buffer and none, plus a tap-level read against a ring sized the way the player sizes it, which must return the audible frames rather than the ones just handed to the backend.

Re-measured after the change, same method as before (`strace -f -e trace=write` on the pty, `Wave`, default buffer): **19.2 FPS**, unchanged from the 19.4 measured before, so the fix costs nothing in refresh rate — it only makes the frames being read the correct ones.

PR #395, opened 2026-08-30
feat(ui): start with the playlist expanded via `expanded` config key and `--expanded`

### LINKED ISSUE #391: Feature request: option to start with the playlist expanded (persist Ctrl+X state)
### What I'd like

A way to have cliamp start in the expanded view — the state `Ctrl+X` toggles — without pressing the key every time. Either a config key (e.g. `expanded = true` in `config.toml`, next to `visualizer` and `repeat`) or a CLI flag would work; a config key is what I'd expect, since the other view preferences already live there.

### Why

On a tall terminal the collapsed view leaves most of the window empty and shows only a few playlist rows, so the first thing I do on every launch is hit `Ctrl+X`. With a 56-track playlist the expanded view is the only one I actually use.

### Current behaviour

As of v1.63.2 the state looks purely runtime — `heightExpanded` in `ui/model/model.go` is toggled by `toggleExpandedView()` (`ui/model/keys.go`) and is never read from or written to the config. `config.Overrides` (`config/flags.go`) has no matching field either, so there is no flag to pass. The only `expand`-ish option I could find is `expand_playlist`, which is about YouTube playlist expansion and unrelated.

If I've missed an existing way to do this, apologies — happy to be pointed at it.

### Workaround (for anyone finding this)

A wrapper that launches cliamp and sends the keystroke a moment later works, on Wayland:

```bash
( sleep 0.6; wtype -M ctrl -k x -m ctrl ) &
exec /usr/bin/cliamp "$@"
```

but it has to pass subcommands (`play`, `next`, `vis`, …) straight through, and it's obviously a hack.

Environment: cliamp 1.63.2 (Arch, `cliamp-bin`), foot on Hyprland/Wayland.

rylos:
Fixes #391.

### The problem

The expanded playlist height that `Ctrl+X` toggles is runtime-only state (`heightExpanded` in `ui/model/model.go`). On a tall terminal the collapsed view shows a handful of rows and leaves the rest empty, so the key has to be pressed on every launch. There was no config key and no flag for it — the only `expand`-ish option, `expand_playlist`, is about YouTube Music playlists.

### The change

Follows the same path `simplified` already takes:

- `Config.Expanded` plus an `expanded` case in the config parser
- `Overrides.Expanded` and `--expanded` / `--no-expanded` on the command line, as a `cli.BoolWithInverseFlag` like every other boolean after #403
- `Model.SetExpanded`, applied at startup next to `SetSimplified`

`SetExpanded` sets the same field `toggleExpandedView` does, and takes no view into account. That matters in simplified mode: `usesSimplifiedLayout()` is transient — it drops as soon as the provider or an overlay takes focus, and those screens do draw a list that `Ctrl+X` expands. `recomputeLayout` already forces `plVisible = 0` while the simplified layout holds, so the height is carried and ignored there rather than blocked, and the flag behaves exactly like the key. `Ctrl+X` is untouched and still collapses the view.

```toml
# config.toml
expanded = true
```

```sh
cliamp --expanded ~/Music
cliamp --no-expanded ~/Music   # collapsed for one session, despite the config key
```

### Tests

`go test ./...` passes. Added:

- `TestLoadExpanded` — the config key in four shapes: `true`, `True`, `false`, absent.
- `TestOverridesApplyExpanded` — the flag override reaches the config.
- `--expanded` and `--no-expanded` rows in `TestInverseBoolFlags`.
- `TestSetExpandedAppliesToSimplifiedProviderLists` — at 100x50 with the provider focused, `--simplified --expanded` takes the list to `layout.bodyRows`, and matches what `toggleExpandedView` produces from the same state.
- `TestSetExpandedIsInertOnTheSimplifiedPlaybackScreen` — the simplified playback screen still reports 0 playlist rows.

### Verified by hand

Arch Linux, foot on Hyprland, 56-track playlist in a 1261x1380 px window: with `--expanded`, and with `expanded = true` in an isolated `CLIAMP_CONFIG_DIR`, the playlist fills the window from the top on startup, exactly as after pressing `Ctrl+X`. Without either, the previous behaviour is unchanged.

### Docs

`config.toml.example` and `docs/configuration.md` get the key, `docs/cli.md` the flag pair.

## Summary by CodeRabbit

* **New Features**
  * Added an option to start the player with the playlist expanded.
  * Configure the initial playlist state in the configuration file or with the `--expanded` command-line flag.
  * The `--no-expanded` option explicitly starts with the playlist collapsed.
  * The existing toggle key remains available during playback.
  * The setting is ignored where no playlist is displayed in simplified mode.

---

**Changes since @gjermundgaraba's review** (rebased onto current `main`):

1. `--expanded` / `--no-expanded` as a `BoolWithInverseFlag`. The override itself was the bug: `overridesFromFlags` did `v := true` instead of `v := c.Bool("expanded")`.
2. Dropped the simplified-mode guard in `SetExpanded`, as described above — it checked the permanent `m.simplified` while the layout checks the transient `usesSimplifiedLayout()`, which is why `Ctrl+X` reached the provider and overlay lists and the flag did not.
3. `expanded` added to `config.toml.example`.
4. The missing blank line in `docs/configuration.md`.
5. "the full list" reworded, and the line that repeated the wrong simplified-mode claim fixed.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -34,6 +34,7 @@ func buildApp() *cli.Command {
 		&cli.BoolWithInverseFlag{Name: "auto-play", Usage: "start playback immediately"},
 		&cli.BoolWithInverseFlag{Name: "simplified", Usage: "simplified playback view (no visualizer or playlist)"},
 		&cli.BoolWithInverseFlag{Name: "help-bar", Usage: "show the key-binding hint bar (? still opens the full keymap)", Value: true},
+		&cli.BoolWithInverseFlag{Name: "expanded", Usage: "start with the playlist expanded (the Ctrl+X state)"},
 		&cli.StringFlag{Name: "provider", Usage: "default provider: radio, podcast, navidrome, lyrion, plex, jellyfin, emby, spotify, qobuz, tidal, soundcloud, mixcloud, netease, yandex, audiobookshelf, abs, yt, youtube, ytmusic"},
 		&cli.StringFlag{Name: "start-theme", Usage: "UI theme name"},
 		&cli.StringFlag{Name: "visualizer", Usage: "visualizer mode"},
@@ -156,6 +157,10 @@ func overridesFromFlags(c *cli.Command) (config.Overrides, error) {
 	if c.IsSet("help-bar") {
 		v := !c.Bool("help-bar")
 		ov.HideHelpBar = &v
+	}
+	if c.IsSet("expanded") {
+		v := c.Bool("expanded")
+		ov.Expanded = &v
 	}
 	if c.IsSet("provider") {
 		v := strings.ToLower(c.String("provider"))
--- a/config/config.go
+++ b/config/config.go
@@ -364,6 +364,7 @@ type Config struct {
 	HideHelpBar      bool                         // hide the key-binding hint bar above the status line
 	HideSettingsPane bool                         // close the settings pane beside the playlist
 	ShowMetadata     bool                         // expand highlighted-track metadata below settings (default false)
+	Expanded         bool                         // start with the playlist expanded (the Ctrl+X state)
 	PaddingH         int                          // horizontal padding for the UI frame (default 3)
 	PaddingV         int                          // vertical padding for the UI frame (default 1)
 	AudioDevice      string                       // preferred audio output device name (empty = system default)
@@ -742,6 +743,8 @@ func Load() (Config, error) {
 				cfg.HideSettingsPane = val == "true"
 			case "show_metadata":
 				cfg.ShowMetadata = val == "true"
+			case "expanded":
+				cfg.Expanded = strings.ToLower(val) == "true"
 			case "audio_device":
 				cfg.AudioDevice = parseString(val)
 			case "initial_directory":
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -91,6 +91,10 @@ simplified = false
 
 # Hide the key-binding hint bar above the status line.
 hide_help_bar = false
+# Start with the playlist expanded, the state Ctrl+X toggles (default: false).
+# The simplified playback screen has no playlist, but its provider and overlay
+# lists do, and they start expanded too.
+expanded = false
 
 # Close the Settings pane beside the playlist (Ctrl+B toggles and saves).
 hide_settings_pane = false
@@ -168,6 +172,12 @@ bar comes back the way you left it. Use `cliamp --no-help-bar` to hide it for
 one session, or `cliamp --help-bar` to show it despite this setting. Simplified
 mode draws neither the hint bar nor a playlist, so it is unaffected by this
 setting.
+
+`expanded = true` starts with the playlist at the expanded height, so the list
+gets every body row the terminal has left instead of the shorter default, and
+`Ctrl+X` is not needed on every launch. The key keeps working and collapses the
+view as before. Start one session with `cliamp --expanded`, or `--no-expanded`
+to start collapsed despite this setting.
 
 List views such as provider browsing, file selection, queues, playlists, search
 results, themes, and keybindings use a content-first layout. This layout replaces
--- a/ui/model/init.go
+++ b/ui/model/init.go
@@ -146,6 +146,18 @@ func (m *Model) SetShowMetadata(v bool) {
 	m.refreshChrome()
 }
 
+// SetExpanded starts the UI in the expanded playlist height, the state the
+// Ctrl+X binding toggles. It sets the same field toggleExpandedView does and
+// takes no view into account: usesSimplifiedLayout() is transient (it drops as
+// soon as the provider or an overlay takes focus), and recomputeLayout already
+// ignores the height while it holds, so a guard here would only make the flag
+// differ from the key.
+func (m *Model) SetExpanded(v bool) {
+	m.heightExpanded = v
+	m.applyHeightMode()
+	m.adjustScroll()
+}
+
 // SetInitialDirectory sets the initial directory for the file browser.
 func (m *Model) SetInitialDirectory(dir string) { m.initialDir = dir }
 
--- a/commands_test.go
+++ b/commands_test.go
@@ -25,6 +25,8 @@ func TestInverseBoolFlags(t *testing.T) {
 		{"--no-simplified", func(ov config.Overrides) *bool { return ov.Simplified }, false},
 		{"--help-bar", func(ov config.Overrides) *bool { return ov.HideHelpBar }, false},
 		{"--no-help-bar", func(ov config.Overrides) *bool { return ov.HideHelpBar }, true},
+		{"--expanded", func(ov config.Overrides) *bool { return ov.Expanded }, true},
+		{"--no-expanded", func(ov config.Overrides) *bool { return ov.Expanded }, false},
 		{"--expand-playlist", func(ov config.Overrides) *bool { return ov.ExpandPlaylist }, true},
 		{"--no-expand-playlist", func(ov config.Overrides) *bool { return ov.ExpandPlaylist }, false},
 		{"--low-power", func(ov config.Overrides) *bool { return ov.LowPower }, true},
--- a/config.toml.example
+++ b/config.toml.example
@@ -71,6 +71,11 @@ eq = [0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
 # section opens full info instead. Closing Settings preserves this preference.
 # show_metadata = false
 
+# Start with the playlist at the expanded height, the state Ctrl+X toggles: the
+# list gets every body row the terminal has left instead of the shorter default.
+# The key keeps working and collapses the view as before.
+# expanded = true
+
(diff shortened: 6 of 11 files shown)


config/config.go
```diff
@@ -695,6 +696,8 @@ func Load() (Config, error) {
 				cfg.Simplified = val == "true"
 			case "hide_help_bar":
 				cfg.HideHelpBar = val == "true"
+			case "expanded":
+				cfg.Expanded = strings.ToLower(val) == "true"
```
rylos, line 747: Thanks — the observation is correct, but it is pre-existing behaviour of the parser rather than something this PR introduces, so I'd rather not fix it here.

`Load()` only skips lines that *start* with `#` (config/config.go:415); inline comments are never stripped, for any key. I checked with a throwaway test on the current branch:

```
config.toml:
  shuffle = true # a
  simplified = true # b
  hide_help_bar = true # c
  expanded = true # d
  buffer_ms = 300 # e
  theme = "Tokyo" # f

result:
  shuffle=false simplified=false hide_help_bar=false expanded=false
  buffer_ms=250 (the default, the Atoi fails)  theme="Tokyo\" # f"
```

So `expanded` behaves exactly like the five booleans next to it — and, thanks to the `strings.ToLower`, it is already slightly more forgiving than `shuffle`/`simplified`/`hide_help_bar`, which compare with `val == "true"` directly.

Fixing only `expanded` would make it the odd one out; fixing it properly means stripping inline comments once, before the switch, so booleans, ints and strings all benefit. That's a parser change with its own blast radius and it deserves its own PR and its own review — I'm happy to open one if @bjarneo wants it.

commands.go
```diff
@@ -157,6 +158,10 @@ func overridesFromFlags(c *cli.Command) (config.Overrides, error) {
 		v := true
 		ov.HideHelpBar = &v
 	}
+	if c.IsSet("expanded") {
+		v := true
+		ov.Expanded = &v
+	}
```
gjermundgaraba, line 164: `--expanded=false` still sets the override to true, and with `expanded = true` in the config there is no `--no-expanded` to start collapsed for one session. since #403 every other boolean that mirrors a config key comes as a `--x` / `--no-x` pair on main, which is also where this file conflicts.

ui/model/init.go
```diff
@@ -122,6 +122,18 @@ func (m *Model) SetHideHelpBar(v bool) {
 	m.refreshChrome()
 }
 
+// SetExpanded starts the UI in the expanded playlist height, the state the
+// Ctrl+X binding toggles. Simplified mode has no playlist to expand, so the
+// request is ignored there, matching what toggleExpandedView produces.
+func (m *Model) SetExpanded(v bool) {
+	if v && m.simplified {
+		return
+	}
```
gjermundgaraba: in simplified mode this returns before touching `heightExpanded`, but provider browsing and the list overlays still draw lists there and their Ctrl+X handlers still expand them, so `--simplified --expanded` does nothing on the only screens where it would show. at 100x50 with the provider focused, Ctrl+X takes the row budget from 24 to 40 and this call leaves it at 24; line 89 of `docs/configuration.md` makes the same claim.

docs/configuration.md
```diff
@@ -85,6 +85,9 @@ simplified = false
 
 # Hide the key-binding hint bar above the status line.
 hide_help_bar = false
+# Start with the playlist expanded, the state Ctrl+X toggles (default: false).
+# Ignored in simplified mode, which has no playlist to expand.
+expanded = false
```
gjermundgaraba, line 97: `config.toml.example` still has `simplified` and `hide_help_bar` but not this key.

docs/configuration.md
```diff
@@ -118,6 +121,10 @@ and gives that row back to the playlist. The full keymap stays available with
 `?`, and `Ctrl+G` toggles the bar for the current session. Start one session
 with `cliamp --no-help-bar`. Simplified mode draws neither the hint bar nor a
 playlist, so it is unaffected by this setting.
+`expanded = true` starts with the playlist at the expanded height, so a tall
```
gjermundgaraba: this runs on from the `hide_help_bar` paragraph without a blank line, so Markdown renders the two as one.

```suggestion

`expanded = true` starts with the playlist at the expanded height, so a tall
```

docs/configuration.md
```diff
@@ -118,6 +121,10 @@ and gives that row back to the playlist. The full keymap stays available with
...
+terminal shows the full list without pressing `Ctrl+X` on every launch. The key
```
gjermundgaraba: "the full list" oversells it: the list gets the rows the terminal has left, 33 of 40 tracks at 100x50 on main and 31 at this base.

gjermundgaraba (changes requested):

commands.go
rylos: Fixed. Rebased onto current `main` and the flag is a `cli.BoolWithInverseFlag` now, like every other boolean after #403.

The actual bug was a line below, in `overridesFromFlags`: it did `v := true` instead of `v := c.Bool("expanded")`, so the override was true whatever the flag said. `--expanded` / `--no-expanded` are now two more rows in the `TestInverseBoolFlags` table.

ui/model/init.go
rylos: You're right, thanks — and the cause is that the two "simplified" notions are not the same one.

`SetExpanded` checked `m.simplified`, the permanent flag, while the layout checks `usesSimplifiedLayout()` (model.go:503), which is transient: it drops as soon as the provider or an overlay takes focus. `toggleExpandedView` has no guard at all, which is exactly why Ctrl+X reached those lists and the flag did not.

So I dropped the guard rather than refining it. It is safe because the consumer already defends itself: while the simplified layout holds, `recomputeLayout` forces `plVisible = 0` (layout.go:91), so the height is carried and ignored instead of blocked, and the flag now sets the same field the key does.

Two tests: at 100x50 with the provider focused, `--simplified --expanded` takes the list to `layout.bodyRows` and matches what `toggleExpandedView` produces from the same state; on the simplified playback screen it still reports 0 rows. I checked the first one fails with the guard put back.

The claim in `docs/configuration.md` is fixed too — it had moved to line 95 after the rebase.

docs/configuration.md
rylos: Added, next to `show_metadata`, with the same commented-out style as its neighbours.

docs/configuration.md
rylos: Applied, thanks.

docs/configuration.md
rylos: Fair — reworded. It now says the list gets every body row the terminal has left instead of the shorter default, with no claim about showing the whole list.

config/config.go
rylos: Re-checked after the rebase onto current `main`: unchanged. `Load()` still skips only lines that *start* with `#`, for every key, so `expanded` behaves exactly like the booleans around it. Stripping inline comments once, before the switch, so booleans, ints and strings all benefit, is a parser change that deserves its own PR — happy to open one if @bjarneo wants it.

docs/configuration.md
```diff
@@ -169,6 +173,12 @@ one session, or `cliamp --help-bar` to show it despite this setting. Simplified
 mode draws neither the hint bar nor a playlist, so it is unaffected by this
 setting.
 
+`expanded = true` starts with the playlist at the expanded height, so the list
+gets every body row the terminal has left instead of the shorter default, and
+`Ctrl+X` is not needed on every launch. The key keeps working and collapses the
+view as before. Start one session with `cliamp --expanded`, or `--no-expanded`
+to start collapsed despite this setting.
```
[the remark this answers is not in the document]
rylos: `site/index.html` does not carry the configuration table, so there is nothing to keep in sync here. It is a landing page: a mock of the UI, the feature list, and a key cheatsheet. It documents no config key and no CLI flag — `simplified`, `hide_help_bar`, `low_power`, `--expand-playlist` and the rest are all absent from it too.

Adding `expanded` there would make it the only option the site mentions, so I'm leaving it out.

config/config.go
```diff
@@ -742,6 +743,8 @@ func Load() (Config, error) {
 				cfg.HideSettingsPane = val == "true"
 			case "show_metadata":
 				cfg.ShowMetadata = val == "true"
+			case "expanded":
+				cfg.Expanded = strings.ToLower(val) == "true"
```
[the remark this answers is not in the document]
rylos: Same finding as #discussion_r3905468921, on the line it moved to after the rebase; the answer is unchanged.

`Load()` skips only lines that *start* with `#`, and it does so for every key — with `shuffle = true # a`, `simplified`, `hide_help_bar`, `buffer_ms` and `theme` all misparse in exactly the same way. `expanded` is not a special case, and fixing only `expanded` would make it the odd one out.

The right fix is to strip inline comments once, before the switch, so booleans, ints and strings all benefit. That is a parser change with its own blast radius and belongs in its own PR — happy to open one if @bjarneo wants it.

PR #404, opened 2026-08-31
feat: allow hiding the key-binding hint bar

### LINKED ISSUE #381: Allow hiding the key-binding hint bar (flag/toggle, auto-hide on idle, or a minimal UI mode)
## Problem

The key-binding hint bar at the bottom of the TUI is always on. There is no way to turn it off.

Once you know the bindings, it stops being information and becomes permanent noise: a full row of text that competes with the track title, the visualizer and the playlist for attention every second the app is open. On short terminal windows it also eats a row that the playlist could use.

`--simplified` already exists for people who want less chrome, but it removes the visualizer and playlist, which is the opposite trade-off: I want the visualizer, I just don't want the cheat sheet.

## Proposal

Any one of these would solve it for me. Listed in the order I'd prefer them, they are not mutually exclusive:

**1. A toggle + flag to hide the hint bar**

- `--no-help-bar` (or `--hide-keybinds`) plus a config key, e.g. `hide_help_bar = true` under the existing UI settings in `config.toml`
- ideally also a runtime keybinding to toggle it on/off, so it can be brought back when you forget a binding, in the same spirit as the other in-app toggles

**2. Auto-hide after a short idle timeout**

Show the hint bar while the user is interacting (scrolling the playlist, changing provider, moving through a menu), then fade it out after ~2 seconds of no input. This is my favourite from a UX point of view: the hints are there exactly when you're navigating and gone while you're just listening.

Ideally the delay is configurable (`help_bar_idle_hide_ms = 2000`, `0` = never hide) so people can pick "always on" or "always off" from the same setting.

**3. A minimal UI mode**

A `--minimal` flag / theme-level option that strips the non-essential rows (hint bar first) but keeps the visualizer and playlist. There already is an internal `layoutMinimal` tier for small terminals; exposing something equivalent as a user choice rather than only as a consequence of window height would cover this.

## Mockups / references

Today (hint bar always present):

```
┌─ CLIAMP ────────────────────────────────┐
│  ▁▃▅▇▅▃▁▃▅▇▆▄▂▁▃▅▇▅▃▁▂▄▆▇▅▃▁▂▄▆         │
│  ▶ Boards of Canada - Roygbiv           │
│  ──────────●───────────────  01:12/02:31│
│  SPD [1x]                    ↓ 3.4 MB   │
│  space play/pause · n next · p prev · / │
└─────────────────────────────────────────┘
```

Wanted (hidden, or auto-hidden after ~2s idle):

```
┌─ CLIAMP ────────────────────────────────┐
│  ▁▃▅▇▅▃▁▃▅▇▆▄▂▁▃▅▇▅▃▁▂▄▆▇▅▃▁▂▄▆         │
│  ▶ Boards of Canada - Roygbiv           │
│  ──────────●───────────────  01:12/02:31│
│  SPD [1x]                    ↓ 3.4 MB   │
└─────────────────────────────────────────┘
```

References in the current code:

- `ui/model/view.go` → `renderHelp()` / `renderTierHelp()`, the hint line that would become conditional
- `ui/model/layout.go` → the existing `layoutMinimal` / `layoutCompact` / `layoutFull` tiers, which already know how to drop rows
- `commands.go` → `--simplified`, `--low-power`, `--start-theme`, precedent for a display-only flag

nfvelten:
Closes #381. Went with option 1 from the issue.

- `hide_help_bar` config key and `--no-help-bar` flag, wired the same way `simplified` is
- `Ctrl+G` toggles it for the session, next to `Ctrl+H`
- hiding drops one row from `fixedRows`, so the playlist gets it back, which was the second half of the issue on short terminals
- the simplified view never draws the bar, so its row budget is left alone

`?` still opens the full keymap, so nothing becomes undiscoverable when the bar is off.

Verified at 80x24 in a real terminal, not just in tests: the hint row disappears and the playlist shows one more track. Tests cover the released row per tier, the simplified case, the omitted section, and `Ctrl+G` through `handleKey`. `make check` passes.

Skipped the idle auto-hide (option 2). It needs an input-activity clock and a redraw timer, and it is a bigger behavioural change than this one. Happy to do it as a follow-up if you want it.

Docs updated in `docs/cli.md`, `docs/configuration.md`, `docs/keybindings.md`, `config.toml.example`, and `site/index.html`.

## Summary by CodeRabbit

* **New Features**
  * Added options to hide the key-binding hint bar with `--no-help-bar` or `hide_help_bar = true`.
  * Added `Ctrl+G` to toggle the hint bar during a session.
  * The full keymap remains available with `?`.
  * Hiding the bar returns its screen space to the playlist and does not affect simplified mode.

* **Documentation**
  * Updated CLI, configuration, keybinding, and feature documentation with the new options.


nfvelten:
Heads-up on overlap with #402 / #403, which landed while I was writing this.

Once `cli.BoolWithInverseFlag` is the convention, a bare hand-written `--no-help-bar` is the odd one out. The consistent shape would be a positive `--help-bar` with the inverse generated for it. Same question for the config key: I used `hide_help_bar` because that is what the issue proposed, but the surrounding keys (`simplified`, `low_power`) read positive, so `help_bar = false` may fit better.

I also touch `commands.go`, `docs/cli.md` and `site/index.html`, so this will likely conflict with #403 textually.

Happy to rebase onto whichever of those lands first and flip to the positive form, or to do it now if you already know that is the direction. Just say which and I will push the update.

nfvelten:
Both review points applied in 6c911cf.

- `mainSections` said it always ends with the hint bar, which stopped being true in this PR. Reworded.
- `docs/configuration.md` and `config.toml.example` now say simplified mode has neither a hint bar nor a playlist and is unaffected by the setting.

On the docstring coverage warning: I documented the functions this PR actually adds or changes behaviour in. The ones still counted as missing are pre-existing and large (`run`, `Load`, `buildApp`, `handleKey`, `overridesFromFlags`), where this diff only adds a few lines inside them. Writing docstrings for those to move the percentage felt like padding the diff rather than improving it, so I left them. Happy to add them if you would rather have the check green.

PR #409, opened 2026-09-01
fix(ui): keep footer a single row so status notifications don't shift the layout

### LINKED ISSUE #408: UI shifts vertically by one row while a status notification is visible (e.g. favoriting a track)
## Summary

Whenever a transient status notification is shown — favoriting a track (`n`), bookmarking (`b`), save feedback, errors, log lines — the **whole UI jumps up by one row** for the notification's TTL (~3s), then drops back down when it disappears. The terminal height itself never changes; the rendered frame grows by one row and gets re-centered.

This isn't specific to favorites: any status message triggers it. Favorite toggles just surface it constantly.

## Root cause

`mainSections` (`ui/model/view.go`) appends the transient notification row as an **extra, conditionally-present line below** the always-present bottom status row:

```go
sections = append(sections, "", m.renderTierHelp(), m.renderBottomStatus())

if includeTransient {
    if line := m.renderTransient(); line != "" {
        sections = append(sections, line)
    }
}
```

The layout only budgets a single footer row (`footerRows: 1` in `ui/model/layout.go`). When a notification is active the frame content is exactly one row taller, so `View` re-centers the frame via `centerFrame` → `padTop = (height - frameH) / 2` — the entire UI shifts up one row and back down after the TTL expires.

## Expected behavior

Showing or clearing a status notification must **not** change the rendered layout/frame height.

tahadx:
Fixes #408

## What

The transient notification row (favorites, bookmarks, save feedback, errors, log lines) was appended **below** the bottom status line as an extra, conditionally-present row. The layout only budgets one footer row (`footerRows: 1`), so an active notification made the frame one row taller and `centerFrame` re-centered the whole UI up by a row for the notification's TTL (~3s), then back down.

The footer is now a single, always-present row: `renderFooterStatus` shows the transient notification when one is active, otherwise the SPD/network status. Frame height is constant, so notifications no longer shift the layout.

## Changes

- `ui/model/view.go`: `mainSections` renders one footer row via `renderFooterStatus` (replaces `renderBottomStatus` + conditional `renderTransient`).
- `ui/model/view_state_test.go`: `TestFooterNotificationDoesNotChangeFrameHeight` — frame height stable with and without an active notification.

## Verification

- `make check` (gofmt + vet + full test suite) passes.

## Test plan

1. `make check`
2. Launch the TUI, play a track, press `n` to favorite/unfavorite — no vertical jump while the notification is visible or after it expires.

## Summary by CodeRabbit

- **Bug Fixes**
  - Notifications and status messages now appear in a dedicated footer row without changing the overall interface height.
  - Footer content remains aligned consistently across standard and simplified layouts.
  - Help information is displayed consistently alongside footer status details.

### DIFF
--- a/ui/model/layout.go
+++ b/ui/model/layout.go
@@ -48,7 +48,7 @@ func (m *Model) recomputeLayout() {
 		panelWidth: max(1, width-2*paddingH),
 		paddingH:   paddingH,
 		paddingV:   paddingV,
-		footerRows: 1,
+		footerRows: 2,
 	}
 	switch {
 	case width < 40 || height < 10:
--- a/ui/model/view.go
+++ b/ui/model/view.go
@@ -271,13 +271,7 @@ func (m Model) mainSections(playlist string, includeTransient, contentFirst bool
 	if !m.hideHelpBar {
 		sections = append(sections, m.renderTierHelp())
 	}
-	sections = append(sections, m.renderBottomStatus())
-
-	if includeTransient {
-		if line := m.renderTransient(); line != "" {
-			sections = append(sections, line)
-		}
-	}
+	sections = append(sections, m.renderFooterStatus())
 
 	return trimTrailingEmpty(sections)
 }
@@ -1098,4 +1092,15 @@ func (m Model) renderBottomStatus() string {
 		return left
 	}
 	return left + strings.Repeat(" ", gap) + right
+}
+
+// renderFooterStatus renders the single always-present footer row. An active
+// transient (error, save activity, status notification, last log line)
+// replaces the speed/network status so the frame height never changes when a
+// notification appears or expires.
+func (m Model) renderFooterStatus() string {
+	if line := m.renderTransient(); line != "" {
+		return line
+	}
+	return m.renderBottomStatus()
 }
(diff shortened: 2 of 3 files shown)


gjermundgaraba (changes requested):
I'm not confident this is the right solution here. the problem is the layout shift, but that is solvable without adding a lot of finicky logic around when to show what in the footer, i think.

Take a look at https://cliamp-pr-409.clankerdrop.com/ and see some examples.
For a bit more context: The jump happens because that reserved row is only rendered when a notification exists, so the frame is one row short of its budget the rest of the time. Rendering the footer row always, blank when idle, fixes the height at every size, covers the simplified view too (which still jumps on this branch), and keeps the SPD row and stream stats on screen instead of replacing them. Your regression test passes unchanged with that change.

tahadx:
> gjermundgaraba

Thanks for the review!

Yeah, yeah… I made some changes.
But we’re working on the Cliamp mobile app, and it has different styles for favorites. I added another PR to match that style here as well.

I’m not sure about using it in the terminal, though—it might make things feel a little too crowded. But on mobile, it’s 🔥!

tahadx:
Closing in favor of a simpler approach

Issue #163, opened 2026-04-02
yt-dlp - can it be installed as a dependency by yay/pacman?


zigmoo:
This error message popped up after I did a ctrl-f search for "rain". (a very very nice feature, btw)

Also on the plus side, great job with the ultra-helpful error message pointing to what the user needs when it's missing!

I'm just wondering if yt-dlp could be installed as one of cliamp's dependencies during normal installation? 
I'm not saying it should be, after all I'm not the designer of cliamp... just a random user. :-)

Thanks in advance,

Z
x.com/zigmoo

<img width="2521" height="1520" alt="Image" src="https://github.com/user-attachments/assets/019283c8-999f-4ec7-9edb-9af9b4d5ea3b" />

electromidz:
I also have same problem in Arhc Linux to install 
If you have problem better to change your mirror list on United State worked for me 
[https://archlinux.org/mirrorlist/](url)
)you can get in this URL and 
set the mirror list on Arch on 
`/etc/pacman.d/mirrorlist
`

PR #422, opened 2026-09-02
feat(yt-dlp): configurable binary path and diagnostics for persistent 403s

megabyte0x:
Two related improvements to selecting and diagnosing yt-dlp playback.

### Select the yt-dlp executable

All yt-dlp call sites resolve the executable through `internal/ytdlbin`, in order:

1. `CLIAMP_YTDLP`
2. `ytdlp_path` in `config.toml`
3. `yt-dlp` from `PATH` (unchanged default)

Place the config key at the top level, before the first `[section]` header:

```toml
ytdlp_path = "~/.local/bin/yt-dlp"
```

This covers playback, duration probing, playlist expansion, downloads, and NetEase cookie extraction. Explicit unusable selections name the binary and selecting setting, without recommending an installation that the selected path would override. Missing and non-executable paths are described accurately. Related: #163.

### Preserve playback diagnostics without hiding the failure

Playback keeps yt-dlp warnings enabled. The one-row TUI prioritizes the final fatal diagnostic; warnings and the captured stderr from failed attempts are available in `cliamp.log` in the configuration directory: `$CLIAMP_CONFIG_DIR` if set, otherwise `$XDG_CONFIG_HOME/cliamp` if set, or normally `~/.config/cliamp`. Existing capture is limited to 64 KiB per process, with a truncation marker. Signal exits caused by stop/seek or other process termination are debug-level rather than normal playback-error entries.

Startup retries classify only the fatal yt-dlp diagnostic. A warning mentioning HTTP 403 cannot turn a permanent `Video unavailable` failure into three attempts. Actual fatal 403s retain the existing three-attempt budget. A monitor goroutine publishes a late process failure into the error latch the UI polls, so a diagnostic reported around PCM EOF is not mistaken for a clean end of track; the audio callback never waits for a process to exit. Exits caused by cliamp's own teardown (stop, seek, an abandoned retry) are cleanup rather than playback failures, decided by an explicit teardown flag instead of an exit code, which a killed process reports differently on POSIX and Windows. A child killed from outside cliamp is still reported as a playback failure.

ffmpeg's PCM now arrives through a pipe cliamp owns rather than `Cmd.StdoutPipe`. `Wait` closes a `StdoutPipe` at process exit, so the end of every track dropped the PCM still buffered in the kernel pipe and failed the next read with `file already closed` instead of EOF — a spurious stream error that triggered a reconnect and could mask the real yt-dlp diagnostic.

There is no automatic version/runtime probe and no probe cache. The guide explains how to run diagnostics manually against the selected binary and covers outdated yt-dlp, JavaScript runtimes, EJS components, and PO tokens.

### Verification

- `make check`
- `make fmt-check vet staticcheck security`
- `go test -count=1 -race ./...`
- Build passed.
- Regression coverage for one-row fatal status, warning preservation in error-level logs, warning-only and multiline 403 false positives, wrapped exit causes, final-error selection, retry budgets, delayed runtime exit diagnostics, process cleanup, and unusable selected paths.
- Diagnostic shell examples exercised against distinct executables for PATH/config/environment selection.

Docs and website updated together.


megabyte0x:
Addressed all three review findings in ef7e24d:

- **Missing pinned binary read as a missing installation.** `ytdlbin.NotFoundError()` now names the binary and the setting that chose it (`yt-dlp not found at /opt/yt-dlp (selected by ytdlp_path)`), while the `PATH` default keeps its previous wording. Used by every yt-dlp call site.
- **Installer ran for a missing explicit selection.** Startup no longer offers to install yt-dlp on `PATH` when `ytdlp_path`/`CLIAMP_YTDLP` pins a binary — the pin keeps precedence, so the install would not be used. It reports the bad path instead. Covered by `TestPinnedYtdlpMissingNotice`.
- **Config example pinned a path.** `ytdlp_path` is commented out in `docs/configuration.md`, so copying the block keeps the documented `PATH` default. The user-local path stays as an override example in `docs/yt-dlp.md`.

`make fmt-check vet staticcheck security` and `go test -count=1 -race ./...` pass locally.

player/ytdl.go
```diff
@@ -291,7 +291,7 @@ func monitorExit(cmd *exec.Cmd, stderr *limitedBuffer, name string) (<-chan erro
 // and returns a streaming PCM decoder. If startSec > 0, ffmpeg -ss is used
 // to skip to the desired position in the input stream.
 func decodeYTDLPipe(pageURL string, sr beep.SampleRate, bitDepth, startSec int) (*ytdlPipeStreamer, beep.Format, error) {
-	if _, err := exec.LookPath("yt-dlp"); err != nil {
+	if _, err := ytdlbin.LookPath(); err != nil {
```
gjermundgaraba, line 382: `decodeYTDLPipe` discards the selector-aware missing-binary diagnostic and always suggests installing the default binary. That cannot fix a missing non-default `CLIAMP_YTDLP` or `ytdlp_path` value because the override keeps precedence. Preserve the selected value/source, and append install guidance only when the selected command is the bare `yt-dlp` PATH lookup. Apply the same rule to the NetEase and resolve guards that currently append install advice.

player/ytdl_test.go
```diff
@@ -61,6 +68,8 @@ cat
 	}
 	t.Setenv("YTDL_ATTEMPTS", attemptsPath)
 	t.Setenv("YTDL_MODE", mode)
+	t.Setenv("YTDL_FAKE_VERSION", time.Now().Format("2006.01.02"))
```
gjermundgaraba: Please clear `CLIAMP_YTDLP` in fake-yt-dlp fixtures. It outranks PATH, so developers with this variable set bypass the fixture and may invoke a real binary/network. I reproduced failures in the player, resolve, and YouTube Music tests with the variable set.

gjermundgaraba (changes requested):
The configurable-binary direction looks right, but please remove the separate persistent-403 health probe. The playback command suppresses yt-dlp’s native stale-version and JS-runtime warnings with `--no-warnings`, then adds roughly 360 lines, including tests, to reproduce them. The final failed attempt already returns captured stderr, so keeping the existing retry and removing `--no-warnings` surfaces those diagnostics without another subprocess that can take up to 10 seconds. It also avoids attaching YouTube-specific advice to 403s from SoundCloud, Bandcamp, Bilibili, and other yt-dlp sources. Please update the matching claims in `docs/yt-dlp.md` and `site/index.html`.

megabyte0x:
Thanks — all three points addressed in 64f5447.

**Removed the health probe.** `player/ytdl_diagnose.go` and its test are gone (-320 lines), and `--no-warnings` is dropped from the playback command. `--quiet` still drops progress chatter but keeps warnings, and I verified the distinction against a real yt-dlp 2026.08.19:

```
$ yt-dlp -q --extractor-args "youtube:player_client=bogus" --simulate <url> 2>&1 >/dev/null
WARNING: [youtube] Skipping unsupported client "bogus"
$ yt-dlp -q --no-warnings --extractor-args "youtube:player_client=bogus" --simulate <url> 2>&1 >/dev/null
(nothing)
```

Warnings go to stderr, so they do not disturb the audio on stdout, and `monitorExit` already returns that stderr with the last failed attempt. The retry loop is unchanged; the third failure now returns the raw error, so no YouTube-specific advice is attached to a 403 from SoundCloud, Bandcamp, or Bilibili, and no second subprocess runs. `docs/yt-dlp.md` and `site/index.html` now describe yt-dlp's own warnings rather than a cliamp probe.

The retry-budget test keeps the coverage: the fixture prints a stale-version WARNING before the 403 and the test asserts it reaches the returned error, plus a `t.Cleanup` that fails if the command ever passes `--no-warnings` again.

**Selector-aware missing-binary errors.** Added `ytdlbin.NotFoundErrorWithAdvice`, which appends install advice only when `Name() == DefaultName`. `decodeYTDLPipe`, the NetEase cookie guard, `resolveYTDLRangePageContext`, and `FetchUserPlaylists` all use it, so a broken `ytdlp_path`/`CLIAMP_YTDLP` now names the selected value and its source with no install advice. Covered by a table test in `internal/ytdlbin` and `TestDecodeYTDLPipeNamesSelectedBinary`.

**Fixtures clear `CLIAMP_YTDLP`.** Done in the player retry fixtures, `TestResolveYTDLBatchCookieSelection`, and both YouTube Music cookie tests. Reproduced your report first — with `CLIAMP_YTDLP` set, `player`, `resolve`, and `external/ytmusic` failed before the change and pass after it.

`make check` and `staticcheck ./...` are clean.

internal/ytdlbin/ytdlbin.go
```diff
@@ -0,0 +1,129 @@
...
+// appended, but only when the bare PATH lookup is in effect. An explicitly
+// selected binary keeps precedence over anything installed on PATH, so telling
+// a user with a broken ytdlp_path or CLIAMP_YTDLP to install yt-dlp sends them
+// after a fix that cannot work.
+func NotFoundErrorWithAdvice(advice string) error {
+	err := NotFoundError()
+	if Name() != DefaultName || strings.TrimSpace(advice) == "" {
+		return err
+	}
+	return fmt.Errorf("%w — %s", err, advice)
```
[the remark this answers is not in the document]
megabyte0x: Skipping: these errors do not reach `main.go`/`run(...)`. `decodeYTDLPipe`, `resolveYTDLRangePageContext`, and `FetchUserPlaylists` fail inside a running TUI, so the string the user reads is the error itself — moving the advice to the process boundary would drop it. The advice was already composed at these call sites before this PR (`"yt-dlp is required — install: %s"`, `"see https://github.com/yt-dlp/yt-dlp#installation"`); the change here only makes it conditional on the bare PATH lookup, which is what the human review asked for. Happy to revisit if @bjarneo wants the wider refactor.

external/netease/provider.go
```diff
@@ -434,8 +435,8 @@ func (p *Provider) ensureCookieHeader(ctx context.Context) (string, error) {
 }
 
 func extractBrowserCookieHeader(ctx context.Context, browser string) (string, error) {
-	if _, err := exec.LookPath("yt-dlp"); err != nil {
-		return "", fmt.Errorf("yt-dlp not found. Install with: %s", ytDLPInstallHint())
+	if _, err := ytdlbin.LookPath(); err != nil {
+		return "", ytdlbin.NotFoundErrorWithAdvice("install with: " + ytDLPInstallHint())
```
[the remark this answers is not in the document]
megabyte0x: Same reason as the sibling comment: `extractBrowserCookieHeader` fails while the TUI is running, so this error text is what the user sees. The `Install with:` phrasing predates this PR; only the "skip the advice when a binary was explicitly selected" rule is new.

docs/yt-dlp.md
```diff
@@ -27,6 +27,42 @@ cliamp search-sc "lofi beats"                  # search SoundCloud
...
+```sh
+CLIAMP_YTDLP=~/.local/bin/yt-dlp cliamp
+```
+
+## Playback fails with HTTP 403
+
+`HTTP Error 403: Forbidden` means the source rejected the media URL yt-dlp
+handed to cliamp. A single 403 is usually transient, and cliamp retries the
+track automatically. When every retry fails, cliamp shows yt-dlp's own output
+for the last attempt, including any warnings it printed. On YouTube the cause
```
gjermundgaraba: These troubleshooting steps are too narrow, and the commands below can inspect a different binary than cliamp uses. Current yt-dlp needs both a supported JavaScript runtime and EJS components for full YouTube support; missing PO tokens can also cause 403s. Please say “common causes include,” mention `yt-dlp[default]`/`yt-dlp-ejs`, link the PO-token guide, and show diagnostics using the binary selected by `CLIAMP_YTDLP` or `ytdlp_path`.

docs/configuration.md
```diff
@@ -93,6 +93,11 @@ theme = "Tokyo Night"
 # Logs are written to ~/.config/cliamp/cliamp.log
 log_level = "info"
 
+# yt-dlp binary to run (default: "yt-dlp" from PATH)
```
gjermundgaraba, line 96: This new selection mechanism makes the existing “Requires `yt-dlp` on `PATH`” statement at line 232 stale. Suggested wording: “Requires an executable yt-dlp, selected by `CLIAMP_YTDLP`, `ytdlp_path`, or PATH.”

gjermundgaraba (changes requested):
have a look at the coderabbit findings + a couple of comments from me :) thanks!

docs/yt-dlp.md
megabyte0x: Rewritten in 4df821a. The section now reads "On YouTube, common causes include:" and lists four:

- outdated yt-dlp (yt-dlp's own `Your yt-dlp version (X) is older than 90 days!` warning),
- no JavaScript runtime (`No supported JavaScript runtime could be found`, links the EJS guide),
- missing EJS components — names [yt-dlp-ejs](https://github.com/yt-dlp/ejs), `pip install -U "yt-dlp[default]"`, notes that official binaries bundle the scripts while distro packages may ship an old one or none, and points at `--remote-components`,
- missing PO token, linking the [PO token guide](https://github.com/yt-dlp/yt-dlp/wiki/PO-Token-Guide).

The diagnostics now target the binary cliamp runs rather than the shell's first `yt-dlp`:

```sh
# $CLIAMP_YTDLP, else the ytdlp_path from config.toml, else yt-dlp on PATH
YTDLP="${CLIAMP_YTDLP:-yt-dlp}"   # or: YTDLP=~/.local/bin/yt-dlp
"$YTDLP" --version
"$YTDLP" -v --simulate 'https://www.youtube.com/watch?v=dQw4w9WgXcQ'
```

with the annotated `-v` header (version, `yt_dlp_ejs-…` in `Optional libraries`, `JS runtimes:`, `[pot] PO Token Providers:`) taken from a real run here.

docs/configuration.md
megabyte0x: Done in 4df821a — line 232 now reads: "Requires an executable yt-dlp, selected by `CLIAMP_YTDLP`, `ytdlp_path`, or `PATH`. See [yt-dlp.md](yt-dlp.md)."

The same statement was stale in four other places, so those got the selection mechanism too: `docs/soundcloud.md` (intro + Requirements), `docs/netease.md`, `docs/mixcloud.md`, and `docs/youtube-music.md`.

resolve/resolve.go
```diff
@@ -673,8 +673,8 @@ func resolveYTDLRangeContext(ctx context.Context, pageURL string, start, end int
 }
 
 func resolveYTDLRangePageContext(ctx context.Context, pageURL string, start, end int, browser ...string) ([]playlist.Track, int, error) {
-	if _, err := exec.LookPath("yt-dlp"); err != nil {
-		return nil, 0, fmt.Errorf("yt-dlp not found in PATH — see https://github.com/yt-dlp/yt-dlp#installation")
+	if _, err := ytdlbin.LookPath(); err != nil {
+		return nil, 0, fmt.Errorf("%w — see https://github.com/yt-dlp/yt-dlp#installation", ytdlbin.NotFoundError())
```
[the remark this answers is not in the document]
megabyte0x: Fixed in 4df821a. `ytdlbin.NotFoundError` now takes the `LookPath` error and keeps it reachable:

```go
type notFoundError struct{ msg string; cause error }
func (e *notFoundError) Error() string { return e.msg }
func (e *notFoundError) Unwrap() error { return e.cause }
```

`errors.Is(err, exec.ErrNotFound)` (PATH lookup) and `errors.Is(err, fs.ErrNotExist)` / `errors.As(err, **exec.Error)` (explicit path) work again at all three call sites, while `Error()` stays the short line the TUI shows instead of pasting in exec's `executable file not found in $PATH`. Covered by `TestNotFoundErrorUnwrapsLookupFailure` and an unwrap assertion in `TestDecodeYTDLPipeNamesSelectedBinary`.

internal/ytdlbin/ytdlbin.go
```diff
@@ -0,0 +1,116 @@
...
+
+// NotFoundError describes an unusable yt-dlp. An explicitly selected binary is
+// named together with the setting that selected it, so a typo in ytdlp_path is
+// not mistaken for a missing installation.
+func NotFoundError() error {
+	name := Name()
+	if name == DefaultName {
+		return errors.New("yt-dlp not found in PATH")
+	}
+	return fmt.Errorf("yt-dlp not found at %s (selected by %s)", name, selectedBy())
```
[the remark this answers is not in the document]
megabyte0x: Fixed in 4df821a. `Name()` is now backed by a `selection()` helper that returns both the resolved command and the setting that chose it, and `NotFoundError` picks the wording from the selector state:

- no selector → `yt-dlp not found in PATH`
- selector holding a bare `yt-dlp` → `yt-dlp not found in PATH (selected by CLIAMP_YTDLP)` / `(selected by ytdlp_path)`
- selector holding a path → `yt-dlp not found at /opt/yt-dlp (selected by ytdlp_path)`

Install advice is still appended in the bare-name case, since a PATH install genuinely fixes it. Regression cases added for both selectors in `TestNotFoundError` and `TestNotFoundErrorWithAdvice`.

external/netease/provider.go
```diff
@@ -434,7 +435,7 @@ func (p *Provider) ensureCookieHeader(ctx context.Context) (string, error) {
 }
 
 func extractBrowserCookieHeader(ctx context.Context, browser string) (string, error) {
-	if _, err := exec.LookPath("yt-dlp"); err != nil {
+	if _, err := ytdlbin.LookPath(); err != nil {
```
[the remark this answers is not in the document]
megabyte0x: This is addressed: the guard returns `ytdlbin.NotFoundErrorWithAdvice(err, …)`, and `NotFoundError` reports `ytdlbin.Name()` plus the setting that selected it (`yt-dlp not found at /path (selected by CLIAMP_YTDLP|ytdlp_path)`), with install advice appended only when the command still resolves through PATH. Verified by `TestExtractBrowserCookieHeaderNamesSelectedYTDLP`.

player/ytdl.go
megabyte0x: Done in 64f5447, extended in 4df821a. `decodeYTDLPipe` now returns `ytdlbin.NotFoundErrorWithAdvice(err, "install: "+YtdlpInstallHint())`, which keeps the selected value and its source and appends install advice only when the command still resolves through PATH. The NetEase cookie guard, `resolveYTDLRangePageContext`, and `FetchUserPlaylists` use the same helper, and `DownloadYTDL` uses the bare `NotFoundError`. 4df821a also handles the case where the selector itself holds a bare `yt-dlp`: the message then says `not found in PATH (selected by CLIAMP_YTDLP)` and keeps the install advice, because a PATH install does fix that one. Covered by `TestDecodeYTDLPipeNamesSelectedBinary`, `TestExtractBrowserCookieHeaderNamesSelectedYTDLP`, and the `internal/ytdlbin` tables.

player/ytdl_test.go
megabyte0x: Done in 64f5447. `installYTDLRetryFixtures` (player), `TestResolveYTDLBatchCookieSelection` (resolve), and both fake-yt-dlp tests in `external/ytmusic` now `t.Setenv(ytdlbin.EnvVar, "")` after prepending the fixture dir to `PATH`.

I reproduced your report first: with `CLIAMP_YTDLP` pointing at a stub binary, `player`, `resolve`, and `external/ytmusic` failed on the parent commit and pass after the change (`CLIAMP_YTDLP=/tmp/real-yt-dlp go test ./... ` is green now). The player fixtures also gained a `t.Cleanup` that fails the test if the playback command ever passes `--no-warnings` again.

internal/ytdlbin/ytdlbin.go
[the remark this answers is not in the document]
megabyte0x: @gjermundgaraba flagging this one for you since it is the last open CodeRabbit finding I have not applied. CodeRabbit withdrew the identical finding on its `external/netease/provider.go` twin after the same argument, so I kept the behaviour consistent across the four guards.

Reasoning: `decodeYTDLPipe`, `extractBrowserCookieHeader`, `resolveYTDLRangePageContext`, and `FetchUserPlaylists` all fail while the TUI is running, so the returned string *is* the user-facing message; there is no `main.go`/`run(...)` boundary left to format it at. The wording also predates this PR (`"yt-dlp is required — install: %s"`, `"see https://github.com/yt-dlp/yt-dlp#installation"`, `"Install with: %s"`) — the only new behaviour is dropping the advice when `CLIAMP_YTDLP`/`ytdlp_path` selects a path, per your review.

4df821a does address the underlying "non-presentational error" concern in the part that matters: the error now carries the `*exec.Error` cause behind `Unwrap`, so `errors.Is`/`errors.As` work regardless of the display text. Happy to move all four messages to a boundary type if you would rather have that.

megabyte0x:
Pushed 4df821a covering your two comments plus the outstanding CodeRabbit findings.

**docs/yt-dlp.md** — the 403 section now says "On YouTube, common causes include:" and lists outdated yt-dlp, no JavaScript runtime, **missing EJS components** ([yt-dlp-ejs](https://github.com/yt-dlp/ejs), `pip install -U "yt-dlp[default]"`, `--remote-components`), and **missing PO token** ([guide](https://github.com/yt-dlp/yt-dlp/wiki/PO-Token-Guide)). Diagnostics now run against the binary cliamp uses (`YTDLP="${CLIAMP_YTDLP:-yt-dlp}"`, or the `ytdlp_path` value), with an annotated `-v` header showing where the version, `yt_dlp_ejs-…`, `JS runtimes:`, and `[pot]` lines appear.

**docs/configuration.md:232** — now "Requires an executable yt-dlp, selected by `CLIAMP_YTDLP`, `ytdlp_path`, or `PATH`." The same stale sentence in `soundcloud.md` (x2), `netease.md`, `mixcloud.md`, and `youtube-music.md` got the selection mechanism too.

**CodeRabbit findings**
- *Preserve the lookup error* (resolve.go:677 + siblings): `NotFoundError` now takes the `LookPath` cause and keeps it behind `Unwrap`, so `errors.Is(err, exec.ErrNotFound)` / `errors.As(err, **exec.Error)` work again while `Error()` stays the short user-facing line. New `TestNotFoundErrorUnwrapsLookupFailure` + an unwrap assertion in `TestDecodeYTDLPipeNamesSelectedBinary`.
- *Preserve the explicit selector* (ytdlbin.go:76): messages are now derived from selector state, so `CLIAMP_YTDLP=yt-dlp` yields `yt-dlp not found in PATH (selected by CLIAMP_YTDLP)` and still keeps install advice, since a PATH install fixes that case. Regression cases added for both selectors.
- *Report the selected executable* (netease provider): already covered; verified by `TestExtractBrowserCookieHeaderNamesSelectedYTDLP`.
- *Move installation guidance to the boundary* (ytdlbin.go:89): left as-is with a note on the thread — these errors surface inside the running TUI, and CodeRabbit withdrew the identical finding on the netease twin. Say the word if you want the wider refactor.

While confirming the `--no-warnings` removal I checked yt-dlp 2026.08.19 source: both `Your yt-dlp version (X) is older than 90 days!` (`update.py:203`) and `No supported JavaScript runtime could be found` (`extractor/youtube/_video.py:2995`) go through `report_warning`, which returns early when `no_warnings` is set — so dropping the flag is exactly what makes them visible.

`make check` + `staticcheck ./...` green, and `CLIAMP_YTDLP=<stub> go test ./...` is green too.

megabyte0x:
@gjermundgaraba could you have a look at this

player/ytdl.go
```diff
@@ -313,14 +313,17 @@ func decodeYTDLPipe(pageURL string, sr beep.SampleRate, bitDepth, startSec int)
 	ytdlArgs := []string{
 		"-f", "bestaudio[protocol=https]/bestaudio[protocol=http]/bestaudio[protocol!=m3u8_native][protocol!=m3u8]/bestaudio/best",
 		"--no-playlist",
+		// --quiet drops progress chatter but keeps warnings on stderr. Do not
+		// add --no-warnings: yt-dlp's own "version is older than 90 days" and
+		// missing-JS-runtime warnings are what explain a persistent HTTP 403,
+		// and monitorExit surfaces this stderr in the failing error.
```
gjermundgaraba: The TUI renders this error in one terminal row, and the playback error is not logged. With a version warning before the fatal diagnostic, the status shows the warning and drops the 403. The same happens with unrelated failures such as `ERROR: Video unavailable`. Please prioritize the fatal diagnostic in the status line and make the full warnings available, for example in the log. Update the matching claims in docs/yt-dlp.md and the site card to describe what users can actually read.

player/ytdl.go
```diff
@@ -410,7 +413,13 @@ func (p *Player) buildYTDLPipeline(pageURL string, startSec int) (*trackPipeline
 		}
 
 		if err := prefillYTDLPipe(decoder); err != nil {
-			if attempt == ytdlPipelineMaxAttempts || !isTransientYTDL403(err) {
+			if !isTransientYTDL403(err) {
```
gjermundgaraba, line 511: `isTransientYTDL403` searches the full stderr text, including warnings. I reproduced three attempts with `WARNING: extractor fallback: HTTP Error 403: Forbidden` followed by `ERROR: Video unavailable`, although the fatal failure is permanent. Please classify the fatal diagnostic, excluding warnings, so a warning mentioning 403 cannot trigger retries.

docs/yt-dlp.md
```diff
@@ -27,6 +27,76 @@ cliamp search-sc "lofi beats"                  # search SoundCloud
...
+```sh
+CLIAMP_YTDLP=~/.local/bin/yt-dlp cliamp
+```
+
+## Playback fails with HTTP 403
+
+`HTTP Error 403: Forbidden` means the source rejected the media URL yt-dlp
+handed to cliamp. A single 403 is usually transient, and cliamp retries the
+track automatically. When every retry fails, cliamp shows yt-dlp's own output
+for the last attempt, including any warnings it printed. On YouTube, common
```
gjermundgaraba: Overstated: the user sees one truncated line, currently the first warning. Same claim in the site card.

docs/yt-dlp.md
```diff
@@ -27,6 +27,76 @@ cliamp search-sc "lofi beats"                  # search SoundCloud
...
+## Choosing which yt-dlp to run
+
+By default cliamp runs the first `yt-dlp` on your `PATH`. Some distributions
+ship a yt-dlp that is months behind upstream, and YouTube stops working with
+old versions quickly. Point cliamp at a current binary instead of shadowing the
+packaged one:
+
+```toml
+# ~/.config/cliamp/config.toml
+ytdlp_path = "~/.local/bin/yt-dlp"
```
gjermundgaraba, line 41: Please say: "Place `ytdlp_path` at the top level, before the first `[section]` header." Appending this example under an existing provider section such as `[soundcloud]` silently ignores the setting. Also, the comment above `YTDLP="${CLIAMP_YTDLP:-yt-dlp}"` describes a config fallback that the shell assignment does not implement. Explicitly tell users with `ytdlp_path` configured to assign that path to `YTDLP`, so the diagnostic command checks the binary cliamp uses.

internal/ytdlbin/ytdlbin.go
```diff
@@ -0,0 +1,153 @@
...
+	case source == "":
+		return &notFoundError{msg: "yt-dlp not found in PATH", cause: cause}
+	case name == DefaultName:
+		return &notFoundError{
+			msg:   fmt.Sprintf("yt-dlp not found in PATH (selected by %s)", source),
+			cause: cause,
+		}
+	default:
+		return &notFoundError{
+			msg:   fmt.Sprintf("yt-dlp not found at %s (selected by %s)", name, source),
```
gjermundgaraba: Minor, non-blocking: an existing non-executable file or directory also produces "not found at". I reproduced both cases. Consider matching the startup notice's "missing or not executable" wording so users can distinguish an unusable selection from a missing installation.

gjermundgaraba (changes requested):
Binary selection works (checked with broken pins and a fake yt-dlp). Two playback issues still need fixing: the one-row TUI error display hides the fatal diagnostic behind the first warning, and the retry classifier treats 403s in warnings as retryable failures. Both are reproduced; details inline, along with documentation corrections and a minor diagnostic wording suggestion.

The PR description still describes the probe and its test, removed in 64f5447.

player/ytdl.go
```diff
@@ -171,27 +174,66 @@ type ytdlPipeStreamer struct {
...
+		// Each monitor reports exactly once on a buffered channel, so two
+		// receives drain both and the goroutine always finishes.
+		for range 2 {
+			select {
+			case e := <-y.ytdlErr:
+				ytdlErr = e
+			case e := <-y.ffmpegErr:
+				ffmpegErr = e
+			}
+		}
```
gjermundgaraba, line 213: this holds an already-known yt-dlp failure until ffmpeg's exit monitor finishes. with a delayed ffmpeg exit after PCM EOF, my process fixture reconnects on base but stops silently on this head; the normal-end `Stop()` sets `closing`, so the saved failure never reaches `StreamErr()`.

player/ytdl.go
```diff
@@ -265,19 +309,63 @@ func (y *ytdlPipeStreamer) Close() error {
...
+func (e *ytdlExitError) Unwrap() error { return e.cause }
+
+// ytdlFatalDiagnostic selects the final ERROR line, not WARNING lines or their
+// continuations. Without a fatal diagnostic, an exit status alone is not enough
+// evidence to retry. Earlier errors can describe failed extractor fallbacks.
+func ytdlFatalDiagnostic(stderr string) string {
+	var diagnostic string
+	for line := range strings.SplitSeq(stderr, "\n") {
+		if text, ok := strings.CutPrefix(line, "ERROR:"); ok {
+			diagnostic = strings.Join(strings.Fields(text), " ")
```
gjermundgaraba, line 336: with `--color always` in yt-dlp's config, its fatal lines start with an ANSI escape, so this prefix check drops the diagnostic. replaying a real colored HTTP 403 returns only `yt-dlp: exit status 1 (see cliamp.log)` and makes one playback attempt on this head, versus three on base.

gjermundgaraba (changes requested):
thanks! a couple more comments and there are some merge conflicts to resolve too

PR #424, opened 2026-09-02
feat: YouTube-style autoplay continuation when the queue runs out

megabyte0x:
## Problem

Play a song from `Ctrl+F` search, let it finish, and playback silently stops.

The search flow calls `playTrackImmediate()`, which appends a **single** track. When it drains, `nextTrack()` → `playlist.Next()` returns `ok=false` (last track, repeat off) → `Stop()` + `clearPlaybackTrack()`. There is nothing left to play, so the player goes quiet.

## Solution

YouTube-style autoplay continuation, opt-in via `autoplay_radio` (default `false`).

When the queue is exhausted and the finished track is a YouTube / YouTube Music video, cliamp fetches the auto-generated **Mix** (`watch?v=<id>&list=RD<id>`) — the same "related tracks" radio YouTube's own autoplay uses — drops entries already in the queue (keyed by video ID, so `www.`/`music.`/`youtu.be` variants match), appends the top 5, and keeps playing. Continuation is recursive: when those run out, the newly finished track seeds the next Mix.

No new yt-dlp plumbing was needed. `resolve` already routes `list=RD...` URLs to `resolveYTDL` with `--flat-playlist`, and `ResolveYTDLBatch` is already exported for the incremental loader.

### Details

- **Early prefetch.** While the last queue track plays, once ≤45 s remain, `preloadNext()`'s no-next branch starts the Mix fetch in the background so the tracks land before the drain and the existing gapless preloader can arm the transition.
- **Failure handling.** A fetch error, or 0 new tracks after dedupe, warns once, stops playback as before, and latches the failed seed in `autoplayFailedSeed` so the tick loop cannot spin up a yt-dlp loop for the same track.
- **Generation guarding.** `requests.autoplay` invalidates in-flight fetches; `beginPlaybackTrack` bumps it and clears the latch whenever a track starts, so a manual action always wins.
- **Unaffected paths.** Repeat one/all (`Next()` never fails), live streams (they reconnect instead of advancing), non-YouTube sources (SoundCloud, local files, …) all behave exactly as before. Autoplay is off unless enabled.

### Usage

```toml
# ~/.config/cliamp/config.toml
autoplay_radio = false   # set true, or press `c` in the player
```

`c` toggles it at runtime in main mode and persists the choice, like `z` (shuffle) and `r` (repeat). The key was free in main mode — the registry only bound `c` inside the queue overlay ("Clear").

Status line: `Autoplay: finding related tracks…` → `Autoplay: added N related tracks`.

## Changes

| Area | Change |
|---|---|
| `resolve/radio.go` | New `YouTubeVideoID()` and `RadioMixURL()` helpers |
| `config/config.go` | `AutoplayRadio` field + `autoplay_radio` parse case |
| `ui/model/autoplay.go` | State, fetch command, eligibility, dedupe/append, prefetch, toggle |
| `ui/model/playback.go` | `nextTrack()` exhaustion branch; autoplay reset in `beginPlaybackTrack` |
| `ui/model/update.go` | `autoplayTracksMsg` handler |
| `ui/model/preload.go` | Early-prefetch hook |
| `ui/model/keys.go`, `command_registry.go` | `c` toggle |
| `main.go` | Config wiring |
| docs, `config.toml.example`, `site/index.html` | User-facing documentation |

## Testing

TDD throughout — 13 new tests in `resolve/radio_test.go` and `ui/model/autoplay_test.go` covering URL parsing, eligibility gating, single-flight fetching, dedupe and the 5-track cap, the exhaustion branch, message handling (append/advance, stale generation, empty result), prefetch windowing, and the `c` toggle including key dispatch and config persistence.

CI parity, run locally:

```
gofmt -l .          # clean
go vet ./...        # clean
staticcheck ./...   # clean
govulncheck ./...   # 0 vulnerabilities affecting this code
go test -race ./... # all pass
```

Manually verified end to end against live yt-dlp: the Mix for a search-played track resolved 20 entries, the seed was correctly deduped, and playback continued onto the next related track. Also smoke-tested in a real TUI session — `autoplay_radio = true` loaded at launch and `c` rewrote the key to `false` in the config file.

Existing end-of-playlist and preload tests are untouched and still pass: they construct models with `autoplayRadio` unset, so the stop-at-end behavior is unchanged when the feature is off.

## Summary by CodeRabbit

* **New Features**
  * Added optional autoplay for YouTube and YouTube Music Mix recommendations when the queue is exhausted.
  * Autoplay adds up to five non-duplicate tracks and continues loading recommendations as needed.
  * Added the `c` key to toggle autoplay during playback; the setting is saved for future sessions.
  * Autoplay applies only to eligible YouTube tracks and is disabled when repeat mode is enabled.
  * Selecting a new track manually removes pending autoplay recommendations.

* **Documentation**
  * Added configuration, keybinding, and YouTube autoplay documentation.

### DIFF
--- a/config.toml.example
+++ b/config.toml.example
@@ -27,6 +27,7 @@ mono = false
 
 # Continue playback with YouTube Mix "related" tracks when the queue runs out
 # (YouTube-style autoplay). Toggle at runtime with the c key.
+# Player only: headless daemon mode (--daemon) ignores this key.
 autoplay_radio = false
 
 # Initial directory for the file browser ('o' key). 
--- a/config.toml.example
+++ b/config.toml.example
@@ -25,6 +25,10 @@ shuffle = false
 # Start with mono output (L+R downmix)
 mono = false
 
+# Continue playback with YouTube Mix "related" tracks when the queue runs out
+# (YouTube-style autoplay). Toggle at runtime with the c key.
+autoplay_radio = false
+
 # Initial directory for the file browser ('o' key). 
 # Supports environment variables ($HOME) and tilde expansion (~/Music).
 # initial_directory = "~/Music"
--- a/config/config.go
+++ b/config/config.go
@@ -351,6 +351,7 @@ type Config struct {
 	Mono             bool
 	Speed            float64                      // playback speed ratio: 0.25–2.0 (default 1.0)
 	AutoPlay         bool                         // start playback automatically on launch (radio streams, CLI tracks)
+	AutoplayRadio    bool                         // continue with YouTube Mix related tracks when the queue runs out
 	SeekStepLarge    int                          // seconds for Shift+Left/Right seek jumps
 	Provider         string                       // default provider: "radio", "podcast", "navidrome", "lyrion", "spotify", "qobuz", "tidal", "plex", "jellyfin", "emby", "audiobookshelf", "soundcloud", "mixcloud", "netease", "yandex", "ytmusic" (default "radio")
 	Theme            string                       // theme name, or "" for ANSI default
@@ -400,6 +401,7 @@ func defaultConfig() Config {
 		VisVolumeLinked: true,
 		Repeat:          "off",
 		AutoPlay:        false,
+		AutoplayRadio:   false,
 		Speed:           1.0,
 		SeekStepLarge:   30,
 		SampleRate:      0,
@@ -696,6 +698,8 @@ func Load() (Config, error) {
 				cfg.Mono = val == "true"
 			case "auto_play":
 				cfg.AutoPlay = val == "true"
+			case "autoplay_radio":
+				cfg.AutoplayRadio = val == "true"
 			case "seek_large_step_sec":
 				if v, err := strconv.Atoi(val); err == nil {
 					cfg.SeekStepLarge = v
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -871,3 +871,26 @@ func TestApplyPlaylist(t *testing.T) {
 		})
 	}
 }
+
+func TestLoadAutoplayRadio(t *testing.T) {
+	t.Setenv("HOME", t.TempDir())
+
+	path := filepath.Join(os.Getenv("HOME"), ".config", "cliamp", "config.toml")
+	if err := os.MkdirAll(filepath.Dir(path), 0o755); err != nil {
+		t.Fatalf("MkdirAll: %v", err)
+	}
+	if err := os.WriteFile(path, []byte("autoplay_radio = true\n"), 0o644); err != nil {
+		t.Fatalf("WriteFile: %v", err)
+	}
+
+	cfg, err := Load()
+	if err != nil {
+		t.Fatalf("Load: %v", err)
+	}
+	if !cfg.AutoplayRadio {
+		t.Error("autoplay_radio = true not parsed")
+	}
+	if defaultConfig().AutoplayRadio {
+		t.Error("AutoplayRadio should default to false")
+	}
+}
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -47,6 +47,7 @@ mono = false
 
 # Continue playback with YouTube Mix "related" tracks when the queue runs out
 # (YouTube-style autoplay). Toggle at runtime with the c key.
+# Player only: headless daemon mode (--daemon) ignores this key.
 autoplay_radio = false
 
 # Initial directory for the file browser ('o' key)
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -45,6 +45,10 @@ shuffle = false
 # Start with mono output (L+R downmix)
 mono = false
 
+# Continue playback with YouTube Mix "related" tracks when the queue runs out
+# (YouTube-style autoplay). Toggle at runtime with the c key.
+autoplay_radio = false
+
 # Initial directory for the file browser ('o' key)
 initial_directory = "~/Music"
 
--- a/docs/keybindings.md
+++ b/docs/keybindings.md
@@ -18,6 +18,7 @@ library commands.
 | `+` `-` | Volume up/down |
 | `]` `[` | Change speed by 0.25x |
 | `m` | Toggle mono |
+| `c` | Toggle autoplay radio (continue with YouTube Mix related tracks) |
 | `Ctrl+J` | Jump to time |
 
 ## Navigation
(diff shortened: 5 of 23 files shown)


megabyte0x:
Thanks — all five findings reviewed against the current code and addressed in `d316d2c`.

**1. Cleanup removed user-owned equivalent tracks (Major, `ui/model/autoplay.go`)** — confirmed and fixed. Keying cleanup off video IDs meant a user-added copy of an autoplay video was deleted too. Replaced the key set with an explicit entry-origin marker: `playlist.Track.Ephemeral`, set on the tracks autoplay appends and on the "play now" track, runtime-only and never persisted (same contract as the existing `DirSourced` flag; `savePlaylist` rebuilds documents field by field). Cleanup now removes only entries carrying that marker, and skips bookmarked ones. Regression tests added for both paths: `TestDiscardKeepsUserOwnedEquivalentTrack` (autoplay entry + user-added `music.youtube.com` copy of the same ID) and `TestPlayTrackImmediateKeepsUserOwnedEquivalentTrack` (play-now entry + user-added `youtu.be` copy). Both fail on the previous commit.

**2. `youtu.be` short-link validation (`resolve/radio.go`)** — confirmed and fixed. `youtu.be/@somechannel` and `youtu.be/<id>/extra` were returned as IDs, and `watch?v=` values were taken verbatim. Added `isYouTubeVideoID` (base64url charset only, so any separator is rejected) applied to both extraction paths, so `RadioMixURL` no longer reports success or builds a Mix URL for them. Near-miss cases added to the table test plus a rejection loop in `TestRadioMixURL`.

**3. Live tracks must not seed a Mix (`ui/model/autoplay.go`)** — fixed. `autoplayEligibleSeed` now also rejects `m.currentPlaybackIsLive(track)`. `preloadNext` already returned early for live tracks, but `nextTrack` did not, so a draining live YouTube stream could have started a fetch. Covered by `TestAutoplayEligibleSeedRejectsLiveStream` (which also asserts `nextTrack` starts no fetch) and `TestAutoplayEligibleSeedRejectsRuntimeDetectedLiveStream` for the `IsLiveStream()` reporter path.

**4. Stale/unwanted results in the handler (`ui/model/update.go`)** — fixed. Before appending, the handler now drops the result when autoplay was turned off mid-flight or when the active playback track is no longer `msg.seed`, clearing `autoplayLoading`/`autoplayAdvance` and returning without touching the queue. The generation check is unchanged and still runs first. Tests: `TestAutoplayTracksMsgDroppedWhenDisabledMidFlight`, `TestAutoplayTracksMsgDroppedWhenSeedNoLongerActive`.

**5. Daemon mode (`main.go`)** — documented as player-only rather than implemented, which was the second option offered. `daemon.nextTrack()` is synchronous with no command loop, so continuation there needs its own async design; that is deliberately out of scope for this PR. `config.toml.example`, `docs/configuration.md` and `docs/yt-dlp.md` now state that `--daemon` ignores `autoplay_radio`.

Verification on `d316d2c`: `gofmt -l .` clean, `go vet ./...`, `staticcheck ./...`, `govulncheck ./...` (0 affecting), `go test -race ./...` all pass.

megabyte0x:
All three findings from the latest review confirmed and fixed in `832ab29`.

**1. `Ephemeral` leaked into the YouTube Music cache (`playlist/playlist.go`)** — real bug, reproduced before fixing. `cachedTrackList.Items` is `[]playlist.Track` marshalled straight into `ytmusic_cache.json`, and `Track` carries no field tags, so the flag was written and restored:

```
snapshot serialized the Ephemeral field: {"scope":…,"items":[{…,"DirSourced":false,"Ephemeral":true,…}]}
cached track restored with Ephemeral = true
```

A restored track would then have been silently removed the next time the user played something. Fixed with `json:"-"` on the field. Added `TestYTCacheDropsEphemeralFlag` in `external/ytmusic`, a full round trip through `snapshot` → `saveSnapshot` → `loadYTCache` → `tracksFresh`, asserting the serialized bytes never contain the field name, the restored flag is false, and other fields survive.

**2. Failure-case assertions in `resolve/radio_test.go`** — fixed. The rejection loop now asserts both halves of the contract, `got == ""` and `ok == false`, in one check.

**3. Non-HTTP(S) schemes (`resolve/radio.go`)** — confirmed and fixed. `url.Parse` happily yields `Host: youtu.be` for `ftp://youtu.be/<id>`, and scheme-relative `//www.youtube.com/watch?v=…` parsed too, so both produced a Mix URL. `YouTubeVideoID` now requires `http` or `https` before any extraction. Regression coverage added for `ftp://` on both host shapes, the scheme-relative form, and a positive `http://` case so plain HTTP keeps working.

Verification on `832ab29`: `gofmt -l .` clean, `go vet ./...`, `staticcheck ./...`, `go test -race ./...` all pass. (Note for anyone reproducing: `mise.toml` pins `go = "latest"`, which now resolves to 1.27.1; a staticcheck binary built with 1.26 reports stdlib compile noise against that toolchain. Ran staticcheck under a matching toolchain for a clean result.)

megabyte0x:
@gjermundgaraba could you please review this.

gjermundgaraba:
this might take some time and we'll probably want to think through the design around this a bit. i really like the idea of having a something like this, but i wonder if there is a more general way to do this that would also allow us to use other providers

hmnd:
Would love for this to come to fruition! The one thing I miss when trying out cliamp

megabyte0x:
> we'll probably want to think through the design around this a bit.

Got it. I will work on it.

gaurabxkc:
@gjermundgaraba on a more general way to do this: #544 already has most of it. It adds a `provider.RadioStarter` interface, `TrackRadio(ctx, trackPath) ([]playlist.Track, error)`, and implements it for Spotify through the client protocol's autoplay context. That is the same station Spotify's own app plays when a queue runs out.

That suggests a provider-agnostic shape for this PR:

- Queue-end continuation asks the provider that owns the finished track for `TrackRadio(seed)`, and appends what comes back, deduplicated against the queue as this PR already does.
- YouTube's Mix (`list=RD<id>`) becomes YouTube's `TrackRadio`. Spotify's comes from #544. Any other provider opts in by implementing the interface.
- `autoplay_radio` stays the single opt-in, and the early prefetch here applies to all of them.

For what it's worth, I have been running Spotify's station as the autoplay source in a fork, through the same client-protocol call #544 uses. For regional music it was far better than any external similarity service: 50 on-genre tracks per seed, where Last.fm returned a handful. Happy to help with the Spotify half once #544 lands, if that is useful.

PR #430, opened 2026-09-04
plex: prompt for library selection in setup wizard

pdwaldrop:
## Summary

Addresses the remaining ask in #206 ("prompt for library selection during setup"). #211 already added a `libraries` key to the `[plex]` config block and made the provider merge albums across all matching libraries, but it required hand-editing `config.toml`. This adds an optional "Music libraries" field to the Plex step of `cliamp setup` so it can be set interactively, using the same comma-separated-list pattern the wizard already uses for Mixcloud's "styles" field. Leaving it blank preserves current behavior (all music libraries load).

## Screenshots / video

This is a TUI, so here's a terminal capture (`tmux capture-pane`) of the live wizard with the new field, filled in:

```
  cliamp setup
  configure remote providers

╭────────────────────────────────────────────────────────────────────────────╮
│                                                                            │
│  Plex Media Server                                                         │
│                                                                            │
│  Stream from your Plex server using an X-Plex-Token.                       │
│  Find a token: https://support.plex.tv/articles/204059436                  │
│                                                                            │
│    Server URL *                                                            │
│    http://192.168.1.10:32400                                               │
│                                                                            │
│    X-Plex-Token *                                                          │
│    •••••••••••••••••                                                       │
│                                                                            │
│  │ Music libraries (optional)                                              │
│  │ Music, Jazz▎                                                            │
│  │ comma-separated names to include; blank loads every music library       │
│                                                                            │
╰────────────────────────────────────────────────────────────────────────────╯
↑/↓ field   enter submit   esc back
```

Submitting this writes:

```toml
[plex]
url   = "http://192.168.1.10:32400"
token = "test-token-abc123"
libraries = ["Music", "Jazz"]
```

## How to test

1. `go test ./cmd/... -run TestPlexSetupBody -v` — covers both the filled-in and blank cases.
2. `cliamp setup` → select Plex Media Server → tab through URL, token, and the new "Music libraries" field → submit → confirm the `[plex]` block in `config.toml` gets a `libraries = [...]` line when the field is non-blank, and no `libraries` key at all when it's left empty.

## Checklist

- [x] `make check` passes
- [x] `docs/` updated (`docs/plex.md` quick-start note); `site/index.html` has no field-level setup content to update for this change

## Summary by CodeRabbit

- **New Features**
  - Plex setup now optionally supports specifying a comma-separated list of music libraries.
  - Generated configuration includes the selected libraries when provided and omits the setting when left blank.

- **Documentation**
[...]

PR #431, opened 2026-09-04
setup: clear stale validation error after "save anyway"

pdwaldrop:
## Summary

Bug found while testing #430 (unrelated Plex feature), reproducible on any provider with a `validate` probe.

Repro: run `cliamp setup`, pick a provider (e.g. Navidrome or Plex), fill in the form with a server that will fail to connect, submit. The wizard shows "Validation failed … Save anyway? [y/N]". Press "y".

- **Expected:** the config section is written, and the result screen shows "⚠ Saved without verification".
- **Actual:** the config section *is* written correctly, but the result screen still shows the original connection-error text with "Press any key to return to the menu," which reads as if the save failed.

Root cause: `persistAndDone(warn bool)` sets `m.resultWarning`, `m.resultText`, and clears `m.awaitingSave` on a successful `saveSection`, but never clears `m.resultErr` (set earlier in `onValidateDone`). `viewResult()` checks `if m.resultErr != nil` before `if m.resultWarning`, so the stale error always wins whenever the user saves after a failed probe.

Fix: clear `m.resultErr = nil` in the success branch of `persistAndDone` (not the `m.saveFailed` branch, which is a genuine failure).

## Screenshots / video

Before (existing behavior — misleading):
```
✗ Validation failed

navidrome: ping.view: ... context deadline exceeded ...

Save anyway?  [y/N]
```
→ press `y` →
```
✗ navidrome: ping.view: ... context deadline exceeded ...

Press any key to return to the menu.
```
(despite the config having been saved)

After this fix, pressing `y` in the same flow instead shows:
```
  Navidrome / Subsonic

  ⚠ Saved without verification

  /tmp/.../config.toml

  Press any key to configure another provider, or q to quit.
```
verified live via `tmux` against an isolated `CLIAMP_CONFIG_DIR`, with the `[navidrome]` block confirmed written to disk in both cases (before and after this fix) — only the result screen was wrong.

## How to test

1. `go test ./cmd/... -run TestSaveAnywayClearsStaleValidationError -v` — new regression test; fails without the one-line fix, passes with it.
2. Manually: `CLIAMP_CONFIG_DIR=/tmp/x cliamp setup`, pick any provider with a validate probe, use an unreachable server, submit, press `y` on "Save anyway?". Confirm the result screen shows "Saved without verification" and `/tmp/x/config.toml` has the section.

## Checklist

- [x] `make check` passes
- [x] `docs/` and `site/index.html` — not applicable; this is an internal state-handling fix with no change to documented behavior or config keys

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed the setup wizard so stale validation errors no longer appear after successfully saving without verification.
  - The result screen now clearly confirms “Saved without verification” while removing outdated connection or validation failure messages.

### DIFF
--- a/cmd/setup.go
+++ b/cmd/setup.go
@@ -1008,6 +1008,7 @@ func (m *setupModel) persistAndDone(warn bool) tea.Cmd {
 	}
 	m.stage = stageResult
 	m.awaitingSave = false
+	m.resultErr = nil
 	m.resultWarning = warn
 	m.resultText = fmt.Sprintf("Saved [%s] section.", spec.section)
 	return nil
--- a/cmd/setup_test.go
+++ b/cmd/setup_test.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"errors"
 	"os"
 	"path/filepath"
 	"runtime"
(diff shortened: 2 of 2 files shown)


pdwaldrop:
Hi @bjarneo — just checking in on this one. Happy to make any changes you'd like; flagging in case it slipped by in the queue. (Opened alongside #430.) CI is pending your approval to run for external contributors.

PR #439, opened 2026-09-05
feat(spotify): load playlist tracks progressively

wundervrc:
## Summary

Opening a large Spotify playlist blocked until every page had been fetched. `Tracks()` paginated `/v1/me/tracks` 50 at a time and the UI waited for the full result, so a large "Liked Songs" library sat on a spinner for about a minute before anything appeared.

This pages tracks into the queue as they arrive. The first page installs the queue and playback is available immediately; each later page appends.

**Request count, spacing and density are unchanged** — paging is still strictly sequential, one page at a time. This adds no rate-limit pressure; the win is that results are visible immediately rather than after the last page.

Mechanically it is a new optional interface, `provider.TrackPager`, asserted at the call site the same way `ResumeTarget` and `CustomStreamer` already are. Only Spotify implements it; every other provider keeps the existing `Tracks()` path unchanged. Pages are driven by a self-chaining `tea.Cmd`, so there are no goroutines or channels and no `tea.Program` handle is needed in `ui/model`.

One related fix came with it. "Liked Songs" is a synthetic ID with no `snapshot_id`, so unlike real playlists it never invalidated its session cache — liking a song elsewhere was invisible until cliamp restarted. It is now revalidated with a single `limit=1` request: `/v1/me/tracks` is ordered `added_at` descending, so an unchanged total plus an unchanged newest entry means nothing was added or removed. Any mismatch, or any error, refetches — the failure direction is stale-free. Repeat visits drop from N requests to 1.

## Screenshots / video

https://github.com/user-attachments/assets/618fa05c-e756-4055-b965-a0603b6ddc89

## How to test

1. Configure a Spotify account with a large Liked Songs library (the effect scales with size; a few thousand tracks makes it obvious).
2. Open the Spotify provider and press Enter on **Your Music**.
3. The queue appears immediately and fills in pages of 50 rather than blocking on the full fetch.
4. Start playback while the list is still growing — the queue stays usable and shuffle mixes new pages into the unplayed tail.
5. Back out with Esc and press Enter again while it is still filling — the list reopens rather than ignoring the key.
6. Leave and re-enter **Your Music** once loaded: it returns from cache after a single revalidation request.
7. Back out mid-load, wait a moment, then re-enter: it picks up where it stopped rather than starting over.
8. Let a track play through to the next one while a large list is still filling: the track that plays is the one the header and MPRIS name.
9. Add or remove a liked song from another Spotify client while a list is loading: the load stops with a notice rather than quietly assembling a list from two different snapshots.
10. Like or unlike a track in another Spotify client, then re-enter **Your Music**: the change is picked up without restarting cliamp.

## Checklist

- [x] `make check` passes
- [ ] `docs/` and `site/index.html` updated for user-facing changes — see below

`make check` passes, and so does the full CI gate locally: `make fmt-check vet staticcheck security` and `go test -count=1 -race ./...`, on Go 1.26.6 to match `go.mod`.

No `docs/` or `site/index.html` change: nothing `docs/` covers has changed — no new keybinding, config key, provider or plugin API. The behaviour differences are the ones named below (a list fills in rather than appearing at once, and gapless is held off while it does), neither of which is documented today. Happy to write them up if you would rather they were.

## The cache invariant

The branch establishes one property worth stating explicitly, because it is what most of the commits are defending: **no writer to `trackCache` can commit a list assembled from more than one library snapshot.**

That matters because a like or unlike prepends, shifting every later offset by one. Pages read either side of such a change splice into a list that is short by one and duplicated by one — and the revalidation probe cannot detect it, since the total and the newest entry both still match. It would be sticky and silent until the next library change.

Both writers are now covered. `TracksPage` pins the total its accumulation started from and refuses any page reporting a different one, so a load that saw the library move never commits — and it stops rather than spending the rest of the library's pages on a result already destined to be discarded. `Tracks()` pins the same way and restarts from a fresh snapshot, bounded at two restarts so a library being actively edited returns an error rather than looping. `Tracks()` is still reachable in both daemon mode and TUI mode — `ui/model/ipc_extended.go` dispatches `provider.tracks` and `provider.load` straight to it — so leaving it unpinned would have left the shared cache poisonable from outside.

Re-entering a list abandoned mid-load resumes the earlier accumulation rather than refetching every page already paid for, which costs one request instead of dozens on a large library. That stitches two separate reads together, so it is only allowed where the pages already held can be proven still valid, and the proof differs by list.

Ordinary playlists use `snapshot_id`, Spotify's own version token, which changes on any edit. The accumulation records the snapshot it began under and one request checks it on re-entry; an unchanged token proves the playlist is untouched. Saved tracks have no snapshot, so they use an unchanged total together with an unchanged head — which is proof only there, because a like lands at position 0 and an unlike moves the total, so no edit can hide from both. An ordinary playlist can be edited anywhere, and a same-total edit below the head but inside the fetched prefix would shift every later offset and stitch the halves together one track short, which is why total and head alone are not enough for them. A missing or unreadable snapshot restarts, which is the safe direction.

Gapless preloading is held off while pages are landing. Each page remixes the upcoming order — and under repeat-all a sequential load wraps to the head at the frontier — so a preload armed mid-load can stop being the next track, and the player advances into a preloaded pipeline on its own thread rather than being told to. It arms again as soon as the order settles, so the cost is no gapless across a boundary falling inside the load window.

One deliberate semantic change: an external `provider.load` over IPC now supersedes an in-flight load rather than racing it. It replaces the queue wholesale, and a paged load would otherwise keep appending its remaining pages onto the list the client just installed.

One behaviour worth naming so it is not mistaken for a bug: opening list B while list A is still loading abandons A's chain silently. A keeps whatever it had, and returns to it for a single request whenever you go back. Backing out with Esc does not abandon a load — it runs to completion — so nothing downloads unless you are looking at it or just were. That is deliberate: a background queue would spend rate-limit budget on lists you may never open again.

One residual is accepted deliberately: unliking one track and liking another between two visits to Liked Songs leaves both the total and the newest entry unchanged, so the revalidation probe serves the cached list. The difference is bounded to that one entry and heals on the next library change.

Saved albums are served whole rather than paged. They come from their own endpoint and are small enough not to need it.

## Notes

Tests are table-driven in `external/spotify/provider_paging_test.go`, using the existing `roundTripFunc` stub. They cover pagination across empty / partial / exact-boundary / multi-page; the revalidation hit and both miss cases (changed total, changed newest entry); rejection of non-contiguous pages; an aborted load leaving nothing cached; the mid-load mutation case for both cache writers, including that a drifted load stops instead of spending the rest of its pages; and the resume path — resuming in a single request through to a complete commit, discarding when the total moved, discarding when the head moved at an unchanged total, resuming a playlist on an unchanged `snapshot_id` and restarting it on a changed one, and serving a saved album whole through its own endpoint rather than splicing an album ID into a playlist-items URL. `ui/model/progressive_tracks_test.go` covers the UI side: the pane unblocking on the first page so a still-filling list can be reopened, later pages appending rather than replacing, a superseded page neither appending nor dispatching, the album-header pin surviving later pages, an IPC load retiring an in-flight paged load, the changed-list notice expiring instead of occupying the permanent error slot while real failures still persist, and the preload guarantees — a stale preload dropped on append, arming suppressed while pages land, and the flag cleared on the terminal page, on error and on provider reset so it cannot strand preloading for the session. The tick test asserts both directions, so the guard cannot pass vacuously.

Two deliberate design points, in case they look odd in review:

- Track totals are stored explicitly rather than derived from `len(tracks)`, because pages skip items with no ID (local/unavailable tracks). Deriving the total would make any library containing local files look permanently changed.
- Partial loads accumulate in a `pending` map and are only promoted to `trackCache` on completion, so navigating away mid-load cannot leave a partial list posing as a complete one. Pages arriving out of order are served to their caller but not accumulated.

The `added_at` ordering that revalidation relies on is undocumented, though consistently observed. It is noted as such in the code, and the failure mode if it ever changes is an extra refetch, never stale data.

Also included, as a separate commit: `coverage.out` added to `.gitignore`. `make coverage` writes it to the repo root, so it shows up untracked after a local CI run. Happy to pull it into its own PR if you would rather keep this one to the provider change.

I have one follow-up queued for this provider — passing `market`, so `is_playable` is actually populated and region-locked tracks are skipped rather than failing at play time. Happy to fold it into this PR if you would prefer one bundle, or keep it separate.

Working on this also turned up a few pre-existing issues that are out of scope here and that I am happy to file separately: `webAPIWithBody` honours a 429 `Retry-After` verbatim, so a long backoff always outlives the caller's context and the retry loop can never complete; `ytdlBatchMsg` appends batches the same way this branch does and leaves a stale gapless preload behind, which is the same defect fixed here for the paged path; and `preloadStreamCmd` and `preloadLocalCmd` are byte-identical.

## Attribution

The code here was written by Claude Code (Opus 5). I directed the work and made the design decisions with it — what to build, what to leave out, and which trade-offs to accept — and a second model, Kimi K3, reviewed every commit. Several of the fixes in this branch exist only because that review caught something: the snapshot pin, the bounded restart in `Tracks()`, and the head comparison guarding resume were all raised there.

Being straight about the limits of that: I have not read the diff line by line myself. What I did do is build it and use it, repeatedly — the re-entry lockout and the full reload on re-entry were both found by hand-testing rather than by either model, and every fix here was re-tested against a real library before this PR, including playing through a track boundary mid-load and mutating the library while a list was still filling. Happy to walk through anything or make changes if a human review turns up something we missed.

## Summary by CodeRabbit

- **New Features**
  - Spotify playlists and saved tracks now load progressively, displaying tracks sooner as additional pages are retrieved.
  - Interrupted page requests can resume, improving navigation through large libraries.

- **Bug Fixes**
  - Improved handling of playlists that change during loading, including stale-result protection and temporary warnings.
  - Prevented outdated loads from replacing newer queue contents or disrupting preloading.
  - Improved saved-track cache validation so library changes are detected more efficiently.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ config.backup.toml
 .env
 .github/workflows/contextowl-changelog.yml
 /gui/
+coverage.out
--- a/external/spotify/provider.go
+++ b/external/spotify/provider.go
@@ -54,9 +54,10 @@ type playlistCache struct {
 // from two snapshots can splice together into a list that is short by one and
 // duplicated by one, which revalidation cannot detect.
 type pendingTracks struct {
-	tracks []playlist.Track
-	want   int
-	total  int
+	tracks   []playlist.Track
+	want     int
+	total    int
+	snapshot string // playlist snapshot_id the accumulation began under; "" for saved tracks
 }
 
 type SpotifyProvider struct {
@@ -519,6 +520,33 @@ func headMatches(accumulated, page []playlist.Track) bool {
 	return true
 }
 
+// snapshotIDLocked returns the snapshot_id last seen for playlistID, or "" if
+// none is known. p.mu must be held.
+func (p *SpotifyProvider) snapshotIDLocked(playlistID string) string {
+	if cached := p.trackCache[playlistID]; cached != nil {
+		return cached.snapshotID
+	}
+	return ""
+}
+
+// playlistSnapshot reads a playlist's current snapshot_id, Spotify's own version
+// token: it changes on every edit, so an unchanged one proves the playlist is
+// untouched -- which an unchanged total and head cannot, since an ordinary
+// playlist can be edited anywhere.
+func (p *SpotifyProvider) playlistSnapshot(ctx context.Context, playlistID string) (string, error) {
+	resp, err := p.webAPI(ctx, "GET", "/v1/playlists/"+playlistID, url.Values{"fields": {"snapshot_id"}})
+	if err != nil {
+		return "", err
+	}
+	var result struct {
+		SnapshotID string `json:"snapshot_id"`
+	}
+	if err := decodeBody(resp, &result); err != nil {
+		return "", err
+	}
+	return result.SnapshotID, nil
+}
+
 // cachedTracksLocked returns a copy of the committed list and its total, if
 // any. p.mu must be held.
 func (p *SpotifyProvider) cachedTracksLocked(playlistID string) (tracks []playlist.Track, total int, ok bool) {
@@ -601,22 +629,43 @@ func (p *SpotifyProvider) TracksPage(playlistID string, offset int) ([]playlist.
 	if next >= total {
 		next = 0
 	}
+
+	// Whether an abandoned accumulation can be resumed may need a request, so
+	// settle it before taking the lock the accumulation is guarded by.
+	resumable := false
+	if offset == 0 {
+		p.mu.Lock()
+		pend := p.pending[playlistID]
+		viable := pend != nil && pend.want > 0 && pend.total == total
+		head := viable && playlistID == savedTracksPlaylistID && headMatches(pend.tracks, page)
+		snapshot := ""
+		if viable && playlistID != savedTracksPlaylistID {
+			snapshot = pend.snapshot
+		}
+		p.mu.Unlock()
+
+		switch {
+		case head:
+			// Saved tracks cannot hide an edit: a like lands at position 0 and an
+			// unlike moves the total, so the head and total together are proof.
+			resumable = true
+		case snapshot != "":
+			current, err := p.playlistSnapshot(ctx, playlistID)
+			resumable = err == nil && current == snapshot
+		}
+	}
+
 	p.mu.Lock()
 	defer p.mu.Unlock()
 	pend := p.pending[playlistID]
 	if offset == 0 {
 		// Re-entering a list abandoned mid-load resumes the earlier accumulation
-		// rather than refetching every page already paid for. Only saved tracks
-		// qualify: an unchanged total and an unchanged head prove nothing moved
-		// there, because a like lands at position 0 and an unlike changes the
-		// total. An ordinary playlist can be edited anywhere, so a same-total
-		// edit below the head but inside the fetched prefix would shift every
-		// later offset and stitch the two halves together a track short.
-		if playlistID == savedTracksPlaylistID && pend != nil && pend.want > 0 &&
-			pend.total == total && headMatches(pend.tracks, page) {
+		// rather than refetching every page already paid for, at the cost of one
+		// request to prove nothing moved in between.
+		if resumable && pend != nil {
 			return slices.Clone(pend.tracks), pend.want, nil
 		}
-		pend = &pendingTracks{total: total}
+		pend = &pendingTracks{total: total, snapshot: p.snapshotIDLocked(playlistID)}
 		p.pending[playlistID] = pend
 	}
 	// A page at an offset this accumulation is not waiting for belongs to a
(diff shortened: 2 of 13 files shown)


wundervrc:
Closing while I verify a couple of things locally — CodeRabbit caught two real defects (saved albums were routed to the wrong endpoint, and resuming an abandoned load was allowed for ordinary playlists where an unchanged total and head prove nothing). Both are fixed on the branch; I want to exercise them against a real library before asking anyone to look. Will reopen once that's done.

wundervrc:
Reopening — the two defects CodeRabbit found are fixed and verified against a real library.

Saved albums were being routed to the playlist-items endpoint and failed to open; they now go through their own endpoint. Resuming an abandoned load was allowed on an unchanged total and head, which proves nothing for an ordinary playlist since it can be edited anywhere — that now uses `snapshot_id`, Spotify's own version token, so an unchanged one proves the pages already read are still valid. Saved tracks keep the total/head check, which is sound only there.

Both are covered by tests, and I re-tested by hand: progressive fill, leaving and returning mid-load, adding and removing tracks while a list loads, playing through a track boundary mid-load, and saved albums.

Issue #436, opened 2026-09-05
Remove theme name from player view


zambetti:
## Problem

The currently selected theme `[Theme: neon-blade-runner]` is displayed inline with playback controls.

This feels out of place in the main player UI. Unlike Shuffle, Repeat, playlist position, etc., the active theme isn't part of playback state and generally isn't information you need while listening.

As it is now, the theme name takes up a valuable prominent chunk of persistent horizontal space with mostly static information.

## Proposal

Remove the active theme name from the main player view. This would keep that row focused on playback/playlist state.

## Mockups / references

<img width="1280" height="844" alt="Image" src="https://github.com/user-attachments/assets/8a856536-f28f-4055-b7cd-101577ac93f8" />

theandrewbailey:
Good idea! I've submitted a PR: https://github.com/bjarneo/cliamp/pull/441

PR #441, opened 2026-09-06
Remove theme name from player view #436

theandrewbailey:
## Summary

Removes the label for the current selected theme. It is static data that isn't all that relevant for the main player view.

## Screenshots / video

CliampNoTheme

## How to test

1. Set a theme.
2. It will not display on the main player view.

## Checklist

- [x] `make check` passes
- [x] (doesn't seem to apply) `docs/` and `site/index.html` updated for user-facing changes

## Summary by CodeRabbit

* **UI Changes**
  * Removed the playlist header’s active theme indicator.

### DIFF
--- a/ui/model/view.go
+++ b/ui/model/view.go
@@ -12,7 +12,6 @@ import (
 	"github.com/bjarneo/cliamp/favorites"
 	"github.com/bjarneo/cliamp/playlist"
 	"github.com/bjarneo/cliamp/provider"
-	"github.com/bjarneo/cliamp/theme"
 	"github.com/bjarneo/cliamp/ui"
 )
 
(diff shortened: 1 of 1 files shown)


RGR09:
Damn, you beat me to it. I was also working on this feature, but implementing it as an optional change, via the use of a `hideTheme` flag, so the existing behavior could be preserved when the flag is disabled. 
I see that your PR removes `themeStr` completly. I think that making it configurable would be useful for users who still want to see the active theme.
I’ve prepared a change that implements that conditional behavior. Would it be useful to open it as a separate PR?

theandrewbailey:
> Damn, you beat me to it. I was also working on this feature, but implementing it as an optional change, via the use of a `hideTheme` flag, so the existing behavior could be preserved when the flag is disabled. I see that your PR removes `themeStr` completly. I think that making it configurable would be useful for users who still want to see the active theme. I’ve prepared a change that implements that conditional behavior. Would it be useful to open it as a separate PR?

Feel free to build on my PR.

PR #442, opened 2026-09-06
feat(jellyfin): restore playback context on launch

TyRichards:
## Summary

- persist the active Jellyfin track, timeline position, and complete source context
- restore the album or other track list with the last track selected when Jellyfin is the default provider
- keep startup silent, including when `auto_play` is configured, until Enter is pressed
- save immediately on track changes and every two seconds so `q`, Ctrl+C, terminal closure, and abrupt exits retain the latest state
- refresh saved Jellyfin stream URLs from current authentication without blocking startup

Restoration only applies when Jellyfin is the default provider and cliamp starts without an explicit file, URL, or playlist.

## Testing

- `make check`
- `go test -race ./internal/resume ./internal/embyapi ./external/jellyfin ./ui/model .`
- manually verified album-context and playback-position restoration through `cliamp-dev`

## Summary by CodeRabbit

- **New Features**
  - Jellyfin playback now remembers the last played track, playback position, and album or track-list context.
  - On the next launch, the saved context and track selection are restored automatically when no files, URLs, or playlists are specified.
  - Resume information is saved when playback starts, periodically during playback, and on normal exit.
  - Restored playback does not automatically start, even when `auto_play` is configured.

- **Documentation**
  - Updated Jellyfin documentation and website descriptions to explain playback restoration behavior.


bjarneo:
Reviewed and updated the branch before merging:

- Refresh restored Jellyfin URLs at engine-open time, including password authentication and preloading, without blocking restoration or changing logical track paths.
- Preserve the correct source list and duplicate occurrence through queue interludes, replacements, and gapless transitions.
- Save confirmed playback positions, not pending seek targets, buffering placeholders, or another gapless track position.
- Add regression tests and keep the Jellyfin docs synchronized with the redesigned site.

HTTP/LAN servers and mixed playlists remain supported; the blanket HTTPS-only and reject-all-mixed-context suggestions would regress existing behavior.

Validation passed: make check; targeted race tests on Go 1.26.6 and Go 1.27.1; go build -trimpath ./...; current staticcheck. The original CI failures were already fixed on main, now included in this branch.

PR #444, opened 2026-09-08
fix(spotify): use the built-in client for catalog reads Development Mode refuses

porchbust:
## Problem

With a personal `client_id`, listing the items of a playlist the user does not own fails:

```
spotify: list tracks: http status 403 Forbidden: {"error":{"status":403,"message":"Forbidden"}}
```

Spotify's Development Mode migration (new apps 2026-02-11, existing apps 2026-03-09) restricts `/v1/playlists/{id}/items` to playlists the user owns or collaborates on. Editorial playlists, a friend's playlist, and anything the user merely follows all return 403. `docs/spotify.md` already noted the restriction; this makes it stop mattering.

## The identity cliamp was discarding

When the user brings their own `client_id`, `interactiveOAuthFlows` already runs a **second** flow against keymaster, because login5 has rejected playback credentials from any other client since August 2026. Keymaster is a first-party client that Spotify does not place in Development Mode, so it is not subject to any of these restrictions.

That flow asked for `playbackOAuthScopes = ["streaming"]` and its refresh token was dropped after minting the librespot session. The one credential that could serve these reads was created and thrown away.

## Change

- Ask the keymaster flow for the full `oauthScopes` instead of `streaming` alone. The built-in-client path (`clientID == DefaultClientID`) already requests exactly these scopes from exactly this client, so this asks Spotify for nothing it does not already grant.
- Persist that flow's refresh token as `catalog_refresh_token` and rebuild the token source on later launches. Failure to re-mint is non-fatal — the session just keeps Development Mode's limits.
- Retry a `403` once under the keymaster token, then surface it if that also fails.

Search deliberately stays on the user's own client, so it keeps the private quota and the existing 10-result Development Mode paging rather than moving onto the shared pool. Only requests Development Mode actually refuses reach the shared client.

### Incidental fix

Refresh-token rotation now goes through `updateCreds`, a mutex-guarded read-modify-write. The previous `webAPITokenSource` closure captured `storedCreds` by value and wrote the whole struct back, so with two rotating token sources each would clobber the other's token. Worth having on its own.

## Compatibility

Credentials stored before this change have no `catalog_refresh_token`, so `catalogSource` stays nil and behavior is unchanged — no failed calls, no forced re-auth on upgrade. Users who want the fallback run `cliamp spotify reset` and sign in once. Users on the built-in client are unaffected: one flow already covers everything, and no second token is stored.

## Tests

`external/spotify/catalog_fallback_test.go`:

- a 403 retries under the keymaster token and succeeds, asserting the exact bearer sequence
- with no keymaster token, the 403 surfaces on the first attempt instead of burning the retry budget
- the keymaster flow carries the full read scopes on both the custom and built-in paths

`gofmt`, `go vet ./...`, and `go test ./...` are clean. `docs/spotify.md` and `site/index.html` updated.

https://claude.ai/code/session_01Ar4wTisgNLjjXvjoVBpTNJ

## Summary by CodeRabbit

* **New Features**
  * Spotify catalog requests that are blocked by Development Mode can retry through the built-in client.
  * Catalog access now supports a separate authorization flow alongside playback authorization.

* **Documentation**
  * Added guidance on catalog-read fallback behavior, required authorization steps, troubleshooting `403 Forbidden` errors, and resetting previously saved credentials.
  * Updated the Spotify provider description to clarify playback and catalog authorization.


gjermundgaraba:
could you share a same-account example where the personal client returns 403 and keymaster returns usable playlist items with this PR’s exact `/items` request and `fields` projection, including whether it still works after restarting cliamp? Our live checks returned 429 twice from keymaster at `/v1/me`, including a recheck with fresh tokens after the first Retry-After interval, so we couldn’t verify that comparison.

Is retrying search and writes under keymaster intentional, or is the intended scope playlist-item reads?

PR #452, opened 2026-09-09
fix(player): fall back to ffmpeg for MPEG-2/2.5 low-rate MP3

cesarho:
## Summary

`go-mp3` (`github.com/hajimehoshi/go-mp3`, the pure-Go decoder `beep`'s `mp3` package wraps — used for all native `.mp3` HTTP/file decoding in cliamp) parses MPEG-2/2.5 Layer III streams — sample rates below the MPEG-1 floor of 32000Hz (22050/24000/16000Hz for MPEG-2, 11025/12000/8000Hz for MPEG-2.5) — **without returning an error**, but produces audibly garbled/distorted PCM. This rate range is common on low-bitrate Icecast/Shoutcast radio, which frequently drops to it for bandwidth-constrained stations.

Root-caused this against a real report of "bad output sound" on `stm2.rthk.hk/radio2` (RTHK Radio 2, a 22050Hz/32kbps MPEG-2 stream). Methodically isolated it by ear:

- Captured raw bytes from the live stream.
- Decoded with `go-mp3` straight to its native 22050Hz — **no resampling at all in the chain** — still garbled. This ruled out resampling, EQ, volume, and the audio device entirely.
- Decoded the identical bytes with `ffmpeg` (`mp3float`) — clean.
- mpv and a browser playing the same URL directly — both clean (they use platform/ffmpeg decoders, not `go-mp3`).

So the defect is specifically `go-mp3` producing wrong output for this MPEG version/rate combination, silently — no decode error to hang a fallback on.

## Fix

This isn't a new code path or a new dependency: `ffmpeg` is already a hard runtime dependency of cliamp, and `needsFFmpeg()` already routes `.m4a`/`.aac`/`.aacp`/`.m4b`/`.alac`/`.wma`/`.opus`/`.webm` and all HLS through it, plus there's already an existing fallback to ffmpeg when a native decoder (`go-mp3`/`wav`/`flac`/`vorbis`) returns an outright error. This PR extends that same, already-exercised fallback to trigger in one more well-defined case.

- `player/decode.go`: adds `isLowRateMP3(ext, sampleRate)` — true for `.mp3` with a sample rate below 32000Hz.
- `player/pipeline.go`: after a successful native decode, if `isLowRateMP3` is true, discards the `go-mp3` decoder's output and routes through the existing ffmpeg fallback path (previously only reachable via a decode error) instead of trusting output known to be wrong for this stream class.

## Testing

- `go build ./...`, `go vet ./...`, full `go test ./...` — all clean across the whole repo.
- New regression tests in `player/decode_test.go`:
  - `TestIsLowRateMP3` — table test covering MPEG-1/2/2.5 rate boundaries and the `.mp3`-only scope.
  - `TestBuildPipelineRoutesLowRateMP3ToFFmpeg` — builds a real pipeline against a small synthesized MPEG-2 22050Hz MP3 fixture (`player/testdata/mpeg2_22050.mp3`, a 1-second sine tone generated with `ffmpeg` — not copyrighted stream audio) and asserts the resulting decoder is `*localFFmpegStreamer`, not the native `go-mp3` path, plus a basic sanity check that it actually produces samples.
- Manually verified end-to-end against the live RTHK stream: built and ran the patched binary, listened to it live and to captured/isolated clips at each pipeline stage — confirmed by ear that the fix resolves it ("very good").

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...`
- [x] New regression tests for the rate detection and the pipeline routing
- [x] Manual end-to-end verification against the live `stm2.rthk.hk/radio2` stream, confirmed by ear

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed playback of low-sample-rate MP3 files that could previously produce garbled audio.
  * These files now use the FFmpeg fallback decoder for reliable playback, whether played locally or streamed online.
  * Preserved stream metadata, such as track titles, during online playback.

* **Tests**
  * Added coverage for low-rate MP3 detection, local and online fallback playback, and metadata preservation.


legacynode:
I independently reviewed and tested this PR locally. The low-rate MP3 detection and FFmpeg fallback look correct, but I found one HTTP-stream regression that should be addressed before merging.

## Issue: ICY metadata is lost for low-rate HTTP MP3 streams

In `player/pipeline.go`, low-rate MP3s are detected after the native decoder has already been created:

```go
if err == nil && isLowRateMP3(ext, format.SampleRate) {
    decoder.Close()
    err = errGoMP3LowRateUnreliable
}
```

For an HTTP URL, the fallback then calls:

```go
p.decodeFFmpegURLStream(path)
```

This makes FFmpeg open the URL directly. It bypasses the `openSource` reader chain, including the ICY reader and its `p.setStreamTitle` callback. As a result, the affected Icecast/Shoutcast streams may play correctly but stop updating `StreamTitle` in the UI.

This matters specifically for the main use case of this PR: low-bitrate HTTP radio streams.

## Suggested fix

For the low-rate HTTP MP3 case, reopen the source with the existing metadata callback and feed that reader to FFmpeg:

```go
replacement, err := openSource(path, p.setStreamTitle)
if err != nil {
    return nil, fmt.Errorf("open fallback source: %w", err)
}

decoder, format, err := decodeFFmpegPipeStream(
    replacement.body,
    p.sr,
    p.bitDepth,
    replacement.live,
)
```

Please make sure the replacement source is closed on all error paths and that the resulting pipeline preserves the replacement source’s:

- `live`
- `prefetch`
- `contentLength`
- byte counter

The existing `decodeFFmpegPipeStream` path was already designed to preserve ICY metadata for FFmpeg-only codecs, so reusing it here should keep the fix narrowly scoped.

## Recommended regression test

Add an `httptest` server that returns:

- a valid MPEG-2 22050 Hz MP3 stream
- an `Icy-Metaint` header
- ICY metadata containing a `StreamTitle`

Then build the pipeline and verify that:

```go
p.StreamTitle()
```

receives the title while FFmpeg decodes the stream through stdin.

## Local verification

The current PR passed locally:

```text
go test ./...
go test -race ./player
go vet ./...
```

The fixture is correctly identified as stereo MP3 at 22050 Hz, and the fallback-routing test passes. I would recommend fixing the HTTP/ICY path and adding the end-to-end metadata regression test before merging.

cesarho:
Thanks — fixed in 7fa19e3.

Added `buildLowRateMP3FFmpegPipeline`: for HTTP URLs it now reopens the source through `openSource(path, p.setStreamTitle)` and feeds ffmpeg via stdin (`decodeFFmpegPipeStream`), same ICY-preserving pattern already used for AAC/Opus, instead of the generic URL-fetch fallback that was dropping ICY metadata. Local files are unaffected.

Added the regression test you suggested (`TestBuildPipelineLowRateMP3HTTPPreservesICYMetadata`) — an `httptest` server serving the MPEG-2 22050Hz fixture with an interleaved `Icy-Metaint`/`StreamTitle` block, asserting the decoder goes through `*ffmpegPipeStreamer` (stdin path) and `StreamTitle()` picks up the title.

`go build`, `go vet`, full `go test ./...`, and `go test -race ./player` all clean. Also manually verified against a live low-bitrate radio stream — audio correct and StreamTitle updates as expected.

cesarho:
@coderabbitai review

@legacynode ready for another look whenever you have a chance — the ICY-preservation fix and regression test are in as of 7fa19e3.

legacynode:
Thanks for the update — this fixes the ICY metadata issue as intended.

I re-ran the local validation:

```text
go build ./...
go vet ./...
go test ./...
go test -race ./player
```

All checks passed. The updated HTTP fallback correctly reopens the source through the ICY-aware reader and feeds FFmpeg via stdin. The local and independent reviews found no further runtime or resource-management issues.

One optional follow-up would be to make the regression test assert explicitly that the `StreamTitle` came from the reopened HTTP request. The current test is already useful and passes reliably; this would only make it more resistant to future regressions.

**From my side, the PR looks good to merge.**

cesarho:
Thanks for the thorough re-review. I took the optional suggestion in 7874425: the test server now sends `Icy-Metaint` and the `StreamTitle` block only on the second request (the reopened, ICY-aware one), so the title can only arrive through the fixed path rather than possibly from the discarded go-mp3 probe connection.

I also checked that it's strictly stronger: it passes on the current code and fails (`StreamTitle()` empty) when run against the pre-fix `pipeline.go`. `go build`, `go vet`, `go test ./...` and `go test -race -count=3 ./player` are all clean. Test-only change, no production code touched.

PR #453, opened 2026-09-09
feat: attach a terminal to a detached cliamp

### LINKED ISSUE #346: TUI attach mode: connect interactive UI to a running --daemon instance
## Problem

`--daemon` (headless) mode is great for systemd/autostart setups: playback survives terminal close, MPRIS works, and everything is scriptable via IPC (`cliamp toggle`, `load`, `status --json`, ...).

But the TUI and daemon are mutually exclusive today — both bind `~/.config/cliamp/cliamp.sock`, so a second instance refuses to start:

> The daemon and TUI share the same Unix socket, so only one cliamp instance can run at a time per user.

That means there's no way to get **background playback + full interactive browsing** at the same time. Concretely: I run the daemon under a systemd user unit so music keeps playing when terminals close, but then I lose search/library browsing, because opening the TUI would require killing the thing that's playing.

This split (thin remote UI ↔ long-lived player process) is exactly how players like `spotify-player` + `spotifyd` or `ncspot` + `librespot` are used, so it's a fairly common expectation.

## Proposal

When a live socket already exists, let `cliamp` (or an explicit flag, e.g. `cliamp attach`) start as a **rendering client** against the running daemon instead of refusing to bind:

- TUI sends IPC commands for user actions it already supports over the socket (play/pause/next/prev/seek/volume/load/queue/shuffle/repeat/EQ/device)
- State that isn't in the current IPC surface would need additions to `status --json` (current playlist contents, provider/browser listings, search results) — or attach-mode could degrade gracefully to only the currently-supported commands
- Daemon stays the single owner of audio output; closing the TUI never affects playback

## Why it looks feasible

The architecture already points this way:

- All CLI subcommands are already thin clients over the socket
- The daemon implements `ipc.Dispatcher`; the TUI dispatches the same logical operations in-process
- So the gap is mainly "render the Bubbletea model from dispatcher responses" rather than new backend functionality

Understandable if full browser-parity attach is a bigger lift (provider browsing/search aren't in the IPC protocol yet) — even a first cut limited to playback controls + queue/status display would cover the systemd-daemon workflow nicely.

Happy to test on NixOS (PipeWire/ALSA) if something lands.

### LINKED ISSUE #377: Headless daemon (--daemon) never reports Audiobookshelf listening progress
## Summary

`cliamp --daemon` never calls `ReportProgress` on the Audiobookshelf provider, so
`mediaProgress` on the server stays empty no matter how long a book plays. Loading
the same book again with `provider.load` always starts at 0:00 instead of resuming,
even though docs/audiobookshelf.md documents automatic resume from the provider pane.

## Environment

- cliamp v2.0.0-rc.1 (linux-amd64 prebuilt binary from Releases)
- Running as `cliamp --daemon`, driven entirely over the v2 IPC socket (no TUI ever
  opened in this session)
- Audiobookshelf provider configured with `url` + `token` in config.toml, confirmed
  reachable (server API returns 200, library visible)

## Repro

```sh
cliamp --daemon &
cliamp remote call provider.load --params '{"provider":"audiobookshelf","playlist":"b:<some-book-id>"}' --wait
cliamp seek 300
sleep 17          # docs say progress reports ~every 15s during playback
cliamp pause
# check server: GET /api/me -> mediaProgress is still []
cliamp stop
cliamp remote call provider.load --params '{"provider":"audiobookshelf","playlist":"b:<same-book-id>"}' --wait
cliamp status --json   # position is 0, not ~300
```

Confirmed against a real Audiobookshelf server (checked `mediaProgress` via `/api/me`
directly): it stays `[]` across the whole sequence above.

## Root cause (found by reading the source)

`Provider.ReportProgress` in `external/audiobookshelf/provider.go:644` is only ever
called from `ui/model/notifications.go:276` — that's the interactive TUI's model
package. Confirmed with:

```sh
grep -rn "ReportProgress(" --include="*.go"
# ui/model/notifications.go:276:  if err := reporter.ReportProgress(track, position); err != nil {
# provider/interfaces.go:124:     ReportProgress(track playlist.Track, position time.Duration) error
```

Nothing in the daemon/headless runtime path calls it, so a `--daemon` session never
reports progress at all, regardless of provider or how long playback runs. This is
separate from #328 (which fixed resume *application* skipping the first fraction of
a second) — that fix is present in this build, but there is no stored position to
apply in the first place because nothing ever writes one outside the TUI.

## Expected

Per docs/audiobookshelf.md ("Progress and resume"), progress should be reported
"about every 15 seconds during playback, and when a track ends" regardless of
whether the session is the TUI or `--daemon`. Headless is the documented, intended
way to drive cliamp from a status bar / script (see docs/headless.md), so resume
silently not working there is surprising given the docs don't call out this gap.

## Suggestion

Wire the same progress-reporting call the TUI model makes into the daemon's
runtime tick (wherever `--daemon` drives playback state), gated the same way
(`provider.TrackPosition`/`ReportProgress`), so both entry points report progress
identically.

ryanrpj:
Closes #346. Closes #377.

`cliamp --daemon` now runs the whole player, just without a terminal of its own, and `cliamp attach` lends it one. `q` hands the terminal back with the music still playing, `cliamp quit` stops the session.

## What this gets you besides attaching

Headless stops being a reduced cliamp, because it is the same runtime now:

- Lua plugins load under `--daemon`. They didn't before, and `docs/headless.md` said as much.
- Gapless preload works headless. Same story: the docs promised gaps between tracks.
- `theme` and `vis` are no longer rejected over IPC, and `cliamp visstream` gets a live spectrum out of a session nobody is watching.
- Provider browsing, search, saved playlists, the EQ, the file browser: whatever the TUI does, a session does, since it is the TUI.
- Listening progress is reported again (#377). `ReportProgress` only ever ran from the model's tick loop, so an Audiobookshelf book played under `--daemon` never got a position on the server.
- One less runtime to keep in sync. `daemon.go` and `daemon_v2.go`, 1834 lines, reimplemented playback, providers, playlist CRUD and V2 dispatch next to `ui/model`. They are gone, and the operation registry no longer has holes to unregister for the daemon.

Nothing changes for a plain `cliamp` in a terminal.

## How it works

Neither of the two shapes discussed in #346, but a third one: keep the running program and hand it a terminal, the way `tmux` does, instead of feeding a second TUI from IPC.

That comes out of `ui/model` not being a renderer of runtime state. It *is* the runtime: the `tickMsg` case owns advance-on-drain, gapless transitions, preload, scrobbling, resume saves and stream reconnects. Routing the model through an `ipc.Dispatcher` would have to switch all of that off in attach mode and mirror it from a snapshot, and would still leave two runtimes carrying the same logic.

The mechanics are Bubble Tea's own public API for handing a terminal over, `ReleaseTerminal` and `RestoreTerminal`, the same one the SSH apps use:

- `session.Host` is the virtual terminal. Output goes to the attached client or nowhere, and input is an `os.Pipe` so the reader can actually be cancelled between attaches.
- The IPC server hands an `attach` request to the host, which acknowledges with a V2 response and then speaks a small frame protocol (input, output, resize, detach) on the same connection. A cliamp that owns a real terminal answers `unavailable`.
- Detached, the model renders no frame and ticks only as often as playback bookkeeping needs, so an idle-but-playing session costs about what the old daemon did: 0.95s against 0.90s of CPU per 10s of radio.

Two notes if you read the host: the client keeps `ONLCR` after `MakeRaw`, because Bubble Tea maps newlines to CRLF whenever the program's input is not a terminal, and the terminal has to be released before the connection closes, since the renderer's teardown is what puts the alternate screen, cursor, window title and theme colors back.

## Limits

- One terminal at a time. A second `cliamp attach` takes the session over and the first client is told why.
- A session renders truecolor and each client downsamples for its own terminal. The depth cannot be renegotiated per attach.
- No key ends a session; its lifetime belongs to whatever started it. `cliamp quit`, `systemctl --user stop` and SIGTERM are all the same graceful shutdown.

## Testing

Client byte streams replayed through a terminal emulator at 200x50 and 160x44 and compared against a normal run: same layout, and the detach stream ends with the same teardown cliamp writes when it exits. New tests in `session/`, `ipc/` and `ui/model/`, and `make check` is clean.

@repparw if you still want to try this on NixOS with PipeWire/ALSA, the branch is on my fork.

## Summary by CodeRabbit

- **New Features**
  - Added detached sessions with full TUI support, allowing playback to continue without an attached terminal.
  - Added `cliamp attach` for connecting, reconnecting, resizing, and taking over a terminal UI.
  - Added `cliamp quit` to stop a running detached session.
  - Detached playback supports continued controls and visualizer updates.

- **User Experience**
  - Pressing `q` detaches from detached sessions while playback continues; it still quits regular sessions.
  - Improved terminal handoff, reconnect behavior, and shutdown handling.

- **Documentation**
  - Updated CLI, remote-control, keybinding, and detached-session documentation.


ryanrpj:
Went through the review. Three of the findings were real bugs and are fixed in fc235d0:

- Both `SetDetachedMsg` branches scheduled a tick, and every tick schedules the next one, so each attach and detach left another chain running forever. Background work grew with every cycle. Nothing is scheduled there now: the next tick reads the new cadence by itself, and rendering does not wait for a tick anyway.
- In the attach client, the size poller received from the same channel the input pump reports through, so whichever won the race decided whether the caller ever saw the input error.
- A failed input write returned quietly, leaving a client that still rendered but forwarded nothing, detach key included. Both exits go through one path now that records why and closes the connection.

The rest: second SIGINT/SIGTERM now exits instead of being buffered and ignored; the handshake read goes through `io.ReadFull` so a zero-byte read cannot spin it; `Detached()` had its doc comment inverted; the `writeTimeout` comment claimed the client is dropped when the render path is actually held for up to five seconds; the startup test no longer calls `t.Fatalf` from a goroutine; and the systemd note dropped `systemctl stop` from the list of exits `Restart=always` undoes, since it does not.

Docstring coverage is up as well, which was the failing pre-merge check.

repparw:
Tested on NixOS (nixos-unstable, PipeWire with the ALSA bridge, Premium, built-in client id). `b9aa7d9` builds over the nixpkgs 2.0.1 derivation. `--daemon` plus `attach` gives the full TUI, tried with radio and Spotify. Gapless, transport IPC, and credential refresh across restarts all worked.

Only quirk, probably pre-existing. A restart lands back on radio until you press `S` again, it does not restore the last provider. `provider = "spotify"` in config covers it. Say the word and I file it separately.

ryanrpj:
> Only quirk, probably pre-existing. A restart lands back on radio until you press `S` again, it does not restore the last provider. `provider = "spotify"` in config covers it. Say the word and I file it separately.

Could you confirm whether you had `provider = "spotify"` in your config during that test?

The provider shown at daemon startup comes from provider in the config file, defaulting to radio when the key is absent. So:

- No provider key: the daemon starts on radio every time. Pressing `S` selects Spotify for that session, but nothing is written back, so the next daemon start is radio again. This matches what you saw, and it's the documented default rather than a regression.
- `provider = "spotify"` set: the daemon starts on Spotify. I tested this path — started the daemon, attached, Spotify was selected; detached and reattached, still Spotify; restarted the daemon entirely, still Spotify.

Attach/detach never re-reads the config in either case. The config is loaded once when the daemon starts and passed into the model at construction; the session layer only lends the daemon a terminal to render into. ***The daemon owns the state and the TUI reads from it, so if you switch provider before detaching, reattaching shows the provider you left it on, not the configured one***. That part I'd defend as correct: detaching shouldn't silently reset where you were.

One thing that is a fair criticism, though: theme, shuffle, repeat and visualizer are persisted back to the config when you change them in the TUI, and provider is not. The reasoning is that the provider pill is navigation state rather than a setting — the same category as which playlist you have open — but the inconsistency is real and I can see why it reads as a bug.

repparw:
meant on daemon restart, but I don't remember if I had Spotify on config yet or had manually navigated to it, so if daemon already shows configured provider on startup you can disregard

bjarneo:
Can you have a look at the merge conflicts?

Issue #413, opened 2026-09-02
Nix flake breaks when go.mod's patch version outruns nixpkgs


ninjapanzer:
Hi — I ran into two Nix build failures this week and figured the pattern behind them was worth reporting, not just the symptoms.

**What I hit**

```
go: go.mod requires go >= 1.26.6 (running go 1.26.5; GOTOOLCHAIN=local)
```

The nixpkgs pinned in `flake.lock` had Go 1.26.5, `go.mod` asked for exactly 1.26.6. Once past that, the `vendorHash` in `nix/package.nix` was also stale from the last dependency bump.

**Why it'll keep happening**

`mise.toml` sets `go = "latest"`, so every `go mod tidy` rewrites the `go` directive to whatever patch release is current that day. nixpkgs is usually a few days behind. Nothing catches it, because your local builds and CI both auto-download the toolchain — the Nix sandbox is the only place that can't, since it has no network and sets `GOTOOLCHAIN=local`.

**Ideas, roughly in order of value**

1. Relax the directive to `go 1.26`. It accepts any 1.26.x, so nixpkgs only has to match the minor version. Add `toolchain go1.26.6` if you want your own machine pinned exactly — Nix ignores that line.
2. Add a `nix build` job to CI on ubuntu-latest and macos-latest. This is the real gap. Nothing exercises the Nix path today, so drift only surfaces when someone tries to install.
3. For the `vendorHash`, either run `nix-update` in CI when `go.mod`/`go.sum` change, or commit `vendor/` and set `vendorHash = null`.
4. `update-flake-lock` on a weekly schedule to keep the nixpkgs pin moving.

Not urgent, and I know Nix probably isn't your daily driver. Happy to send patches for any of it if you'd rather not spend the time.

Issue #142, opened 2026-03-26
Fails to build on NixOS


eggy111:
The package is broken in nixpkgs, so I downloaded the repo to build it from source, and I still receive an error while trying to build it.  It says "pkg-config": executable file not found in $PATH, for vorbis, and oto/v3, go-librespot/flac and go-librespot/output.  Sadly I don't know what these mean.  I will paste the package.nix file I am using to build it with, hopefully I didn't do anything wrong! Thanks

`  1 │ # First, the named parameters.
   2 │ {
   3 │   # you almost always want to depend on lib
   4 │   lib,
   5 │ 
   6 │   # these are Nixpkgs functions that your package uses
   7 │   buildGoModule,
   8 │   fetchFromGitHub,
   9 │ 
  10 │   # more dependencies would go here...
  11 │ }: # this means end of named parameters
  12 │ 
  13 │ # Now, the definition of your package.
  14 │ # This should be something that produces a derivation, not
  15 │ # a string or a raw attribute set or anything else.
  16 │ # buildGoModule is a function that returns a derivation, so
  17 │ # you want `buildGoModule ...` here, not `{ pet = ...; }` here;
  18 │ # the latter is an attribute set.
  19 │ 
  20 │ buildGoModule rec {
  21 │   pname = "cliamp";
  22 │   version = "1.27.8";
  23 │ 
  24 │   src = fetchFromGitHub {
  25 │     owner = "bjarneo";
  26 │     repo = "cliamp";
  27 │     rev = "v${version}";
  28 │     hash = "sha256-SE8uUdA3u+bm3A65J2nMyG+17NCaq5mQgji7eYUQKd8=";
  29 │   };
  30 │ 
  31 │   # this hash is updated from the example, which seems to be out of date
  32 │   vendorHash = "sha256-UMDCpfSGfvJmI+sImaFzgZpLNaLMgEnmGCqERwPokHM=";
  33 │ 
  34 │   meta = {
  35 │     description = "cliamp - Terminal music player inspired by winamp ";
  36 │     homepage = "https://github.com/bjarneo/cliamp";
  37 │     license = lib.licenses.mit;
  38 │     maintainers = with lib.maintainers; [ bjarneo ];
  39 │   };
  40 │ }`
sorry if bad formatting

PR #459, opened 2026-09-09
fix(nix): refresh vendorHash so the flake builds again

noosxe:
## Problem

`nix build github:bjarneo/cliamp` currently fails on `main` (93a4ac6) at the `go-modules` fixed-output derivation:

```
error: hash mismatch in fixed-output derivation '...-cliamp-<rev>-go-modules.drv':
         specified: sha256-rtwUWbft5XGEbuBCn0OMCn4TS5Ul+UXJNIqNOzXfU+M=
            got:    sha256-d/ENFm9b1DkIir1lz50VVX1pvuQpwPUVlA5XOC7Jj5o=
```

The `vendorHash` committed in #414 no longer matches the module set the current `go.mod`/`go.sum` produces. Since CI has no Nix job, this went unnoticed — same class of breakage as #142 and #413.

## Fix

Refresh `vendorHash` to the value Nix computes from the actual dependency closure (one line).

## Verification

Built on `x86_64-linux`:

- with this repo's own `flake.lock` nixpkgs pin: ✅
- against current `nixos-unstable` (d6524aa, via `inputs.nixpkgs.follows`): ✅ — both produce the same `go-modules` hash, so the value is stable across the two nixpkgs revisions.

## Follow-up suggestion

A minimal CI job running `nix build .#cliamp` (one runner, `ubuntu-latest` + install Nix) would catch stale `vendorHash` at PR time instead of after users hit it.

## Summary by CodeRabbit

* **Chores**
  * Updated package build metadata to match the current set of vendored Go dependencies.
  * No user-facing functionality or behavior changes.

### DIFF
--- a/nix/package.nix
+++ b/nix/package.nix
@@ -38,7 +38,7 @@ buildGoModule {
   inherit version;
 
   src = lib.cleanSource ../.;
-  vendorHash = "sha256-rtwUWbft5XGEbuBCn0OMCn4TS5Ul+UXJNIqNOzXfU+M=";
+  vendorHash = "sha256-d/ENFm9b1DkIir1lz50VVX1pvuQpwPUVlA5XOC7Jj5o=";
 
   nativeBuildInputs = [
     makeWrapper


legacynode:
Verified independently on native Apple Silicon.

Environment: macOS 26.5.2, arm64 (M4), Determinate Nix 3.16.0 / Nix 2.33.3.

At commit `6202a4f8fbf909f9cd962a5a225a571f72c52fd0` I ran:

- `nix flake check --all-systems -L`
- `nix build .#default -L`
- `./result/bin/cliamp --version`
- `nix run . -- --version`
- `nix develop .#default --command ...`

Results: all commands passed; the package build completed its full Go `checkPhase` successfully, and both the built binary and `nix run` printed `cliamp version 6202a4f`. The corrected `vendorHash` works on `aarch64-darwin` as well as the Linux verification in the PR description.

No issue found in this one-line fix.

PR #460, opened 2026-09-10
test(jellyfin): assert standard Authorization header with MediaBrowser scheme

setiapam:
Rebased onto `main` following the merge of #471.

### Changes
Narrowed the scope to keep test coverage for standard `Authorization` headers:
- Added explicit assertions in `TestJellyfinAuthHeaderScheme` verifying `Authorization: MediaBrowser ...` with token.
- Added explicit assertions in `TestJellyfinAuthenticatesWithPassword` verifying `Authorization: MediaBrowser ...` on login and subsequent library view requests.
- Added test function docstrings.

All unit tests in `internal/embyapi` pass cleanly.

## Summary by CodeRabbit

- **Bug Fixes**
  - Authentication requests now consistently use the `MediaBrowser` authorization scheme and include the access token when retrieving views.
  - Improved verification helps ensure authenticated Jellyfin requests are accepted reliably.


gjermundgaraba:
Thanks for the fix and the regression tests! We've merged #471, which covers the authentication-header change along with related streaming and API-key fixes.

Your explicit checks for the standard `Authorization` header still fill a gap in our tests. Would you be willing to rebase this PR onto main and narrow it to those test additions? We'd like to keep that coverage.

setiapam:
Rebased onto `main` and narrowed the PR strictly to the regression test assertions in `internal/embyapi/client_test.go` as requested. All unit tests pass cleanly.

gjermundgaraba (approved):

PR #463, opened 2026-09-10
Fix Jellyfin API compatibility

diegodorado:
## Summary

* Fix Jellyfin API compatibility for library discovery and authentication.
* Use Jellyfin's `/Library/MediaFolders` endpoint instead of the Emby `/Users/{id}/Views` endpoint.
* Update Jellyfin authentication to use the `Authorization: MediaBrowser ...` header format and avoid requiring a Jellyfin user ID.
* Add dialect-specific library path handling while keeping the common client logic shared with Emby.
* Fix `q` handling in the setup result screen so it properly quits instead of returning to the setup menu.

## Testing

* Verified Jellyfin authentication and library access against Jellyfin 10.10.7.
* Verified album and track retrieval through the Jellyfin API.

## Summary by CodeRabbit

- **New Features**
  - Improved Jellyfin compatibility for sign-in, server detection, and music library browsing.
  - Added support for Jellyfin’s library paths and authorization requirements.
  - Library loading now works consistently across supported media servers.

- **Bug Fixes**
  - The setup wizard now exits when you press `q` on the result screen, including after validation failures.
  - Pressing Enter, Escape, or Space returns to the provider selection menu as expected.
  - Updated prompts clearly indicate the available quit option.

### DIFF
--- a/cmd/setup.go
+++ b/cmd/setup.go
@@ -1035,7 +1035,9 @@ func (m *setupModel) resultKey(msg tea.KeyPressMsg) (tea.Model, tea.Cmd) {
 	}
 
 	switch msg.String() {
-	case "enter", "esc", "q", " ":
+	case "q":
+		return m, tea.Quit
+	case "enter", "esc", " ":
 		m.stage = stageMenu
 		return m, nil
 	}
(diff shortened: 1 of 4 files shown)


gjermundgaraba:
Thanks for investigating this and putting together a fix! We've merged #471, which addresses Jellyfin authentication and streaming while preserving user-specific library views, so we're closing this PR as superseded.

The setup-screen `q` fix is separate and still looks useful. Would you be willing to submit that as a focused PR against current main? We'd be happy to review it.

diegodorado:
ok, will resubmit a focused PR, thanks

Issue #451, opened 2026-09-09
Cliamp 2.2.0 cannot connect to Jellyfin 10.10.7 or 12


diegodorado:
## What happened

Cliamp 2.2.0 cannot connect to a Jellyfin server. `cliamp setup` fails with `400`/`401` responses and reports `validation failed`.

I initially suspected this was caused by Jellyfin 12's removal of legacy authentication, but the same problem occurs with a freshly installed Jellyfin 10.10.7 server.

I tested all three authentication methods supported by the cliamp setup:

* Jellyfin API token
* Jellyfin username/password
* Jellyfin user access token

**None of them work.**

The Jellyfin server itself is reachable and healthy. The same credentials can successfully authenticate using Jellyfin's HTTP API directly.

For example, a user access token successfully authenticates against `/Users/Me`:

```bash
curl -i \
  -H 'Authorization: MediaBrowser Client="curl", Device="curl", DeviceId="curl-test-1", Version="1.0.0", Token="ACCESS_TOKEN"' \
  http://localhost:8096/Users/Me
```

This returns HTTP 200.

An API token also successfully authenticates against a server-level endpoint:

```bash
curl -i \
  -H 'Authorization: MediaBrowser Client="curl", Device="curl", DeviceId="curl-test-1", Version="1.0", Token="API_TOKEN"' \
  http://localhost:8096/System/Info
```

This returns HTTP 200.

So the Jellyfin server, network connection, and credentials are working; the failure appears to be in cliamp's Jellyfin integration.

## Steps to reproduce

1. Install Jellyfin 10.10.7 on macOS using the official DMG.
2. Start Jellyfin and configure a user and music library.
3. Run `cliamp setup`.
4. Configure the Jellyfin server URL.
5. Try API token authentication, username/password authentication, and a user access token.
6. Cliamp fails validation with HTTP `400`/`401` in all cases.

## Expected behavior

Cliamp should authenticate with the Jellyfin server and successfully validate the provider configuration.

## Screenshots / video

N/A

## Environment

* cliamp version (`cliamp --version`): 2.2.0
* OS: macOS (Apple Silicon)
* Audio backend (PipeWire / PulseAudio / ALSA): N/A
* Provider(s) involved: Jellyfin
* Jellyfin version: 10.10.7

diegodorado:
I fixed it here: https://github.com/bjarneo/cliamp/pull/463

gjermundgaraba:
[...]

PR #471, opened 2026-09-11
fix(jellyfin): restore auth and streaming on Jellyfin 12

heggrik:
## Summary

Fixes Jellyfin authentication and streaming on Jellyfin 12 (and 10.11.2+), where legacy auth handling broke both the connection check (`Ping`) and track playback.

## What changed

- **Auth headers**: Jellyfin 10.11.2+/12 read the client identity from the standard `Authorization` header on `/Users/AuthenticateByName`, while older servers read `X-Emby-Authorization`. The Jellyfin dialect now sends both (`MediaBrowser ...` scheme), keeping the legacy header for older servers.
- **Stream URLs**: Item download URLs now carry both the `ApiKey` and the legacy `api_key` query parameters, so buffered stream requests that carry no headers work on any server version.
- **User-id discovery**: `/Users/Me` returns 400 for server-level API keys because a key isn't owned by a user. Discovery now falls back to `/Users` and prefers the user matching the configured username.
- **Ping**: `Client.Ping` falls back to `/Users` only when the dialect is Jellyfin, both credentials are empty, and the ping returns HTTP 400. The fallback is anchored on a new typed `httpError` that preserves the status code while keeping raw response bodies out of error strings; unrelated failures and all Emby paths return the original error unchanged.

## Notes

- Emby is unaffected: the `/Users` fallback is gated to the Jellyfin dialect, and Emby's ping endpoint (`/System/Info`), auth scheme, and user-id discovery are untouched.
- Jellyfin's documented "Token is not owned by a user." message is sent as an empty response body, so the fallback matches on status 400 rather than body text.

## Testing

- Unit tests covering the API-key ping fallback (success and invalid-token paths), the non-400 / unrelated-error cases that must not fall back, Emby ping failures never falling back, and both auth query parameters in stream URLs.
- `make check` passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Jellyfin authentication compatibility across supported authorization formats.
  * API-key connections now validate more reliably, including fallback user discovery where appropriate.
  * Stream URLs support both current and legacy API-key formats.
  * Improved connection error reporting with clearer status and response details.
  * Preserved specific authentication and library-access failures instead of applying inappropriate fallback behavior.

### DIFF
--- a/external/jellyfin/provider_test.go
+++ b/external/jellyfin/provider_test.go
@@ -188,7 +188,7 @@ func TestProviderRestoreTrackDefersAuthenticationUntilSourceResolution(t *testin
 	if err != nil {
 		t.Fatalf("ResolveSource() error: %v", err)
 	}
-	if want := "https://jf.example.com/Items/track-1/Download?api_key=new-token"; source != want {
+	if want := "https://jf.example.com/Items/track-1/Download?ApiKey=new-token&api_key=new-token"; source != want {
 		t.Fatalf("ResolveSource() = %q, want %q", source, want)
 	}
 	if got.Path != oldURL || got.Title != "Song" || got.Meta(provider.MetaJellyfinID) != "track-1" || !got.Stream {
--- a/external/jellyfin/provider_test.go
+++ b/external/jellyfin/provider_test.go
@@ -164,6 +164,9 @@ func TestProviderRestoreTrack(t *testing.T) {
 	}
 }
 
+// TestProviderRestoreTrackDefersAuthenticationUntilSourceResolution verifies
+// that restoring a saved stream URL performs no startup requests and defers
+// password authentication until source resolution.
 func TestProviderRestoreTrackDefersAuthenticationUntilSourceResolution(t *testing.T) {
 	p := newProvider(NewClient("https://jf.example.com", "", "", "user", "password"))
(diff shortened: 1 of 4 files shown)


heggrik:
@coderabbitai resume

heggrik:
this PR fixes issue [#451](https://github.com/bjarneo/cliamp/issues/451)

gjermundgaraba:
amazing @heggrik! testing and verifying now before i do a review pass on this 🙏

gjermundgaraba (approved):

gjermundgaraba:
Verified at `31cb88a16eb3d4904f2560bb977310ac7f19e502`:

- Jellyfin 10.10.7 and 12.0.0: password, user-session token and server API-key authentication, library browsing and audio playback work, including Jellyfin 12 with legacy authentication disabled.
- Restricted users retain their own library views. API keys select the configured username/user ID; without either, they select the first returned user. A nonexistent configured username fails discovery, and invalid tokens/passwords are rejected.
- API-key connection validation and user discovery succeed where the base failed. Existing Jellyfin 10.10.7 password/session playback and Emby 4.9.3.0 password/API-key playback remain working.
- Full Go tests, vet, formatting, build and targeted shared-client/Jellyfin/Emby race checks pass.

No blocking issues found. Interactive setup, end-to-end resume/gapless and server-side playback reporting were not verified.

PR #485, opened 2026-09-13
feat(config): add cookies_file for yt-dlp providers

gjermundgaraba:
## Summary

Adds `cookies_file` to `[ytmusic]`, `[soundcloud]`, `[mixcloud]` and `[netease]`: a Netscape-format `cookies.txt` used instead of a browser profile, so signed-in playback works where no browser is installed (servers, containers, SSH). `cookies_from` keeps working; the file wins when both are set. A leading `~` is expanded.

- `internal/ytdlcookies` stores a `Source{Browser, File}` per host and renders the yt-dlp flags itself (`--cookies-from-browser` / `--cookies`); the repeated flag-building at each yt-dlp call site is gone.
- yt-dlp **rewrites its `--cookies` jar on exit** (and fails with a traceback if the file is read-only). Each yt-dlp process therefore gets a private temp copy, removed once the process exits: the user's export is never modified, read-only mounts work, and concurrent invocations (playback + gapless preload + duration probe) don't race on one jar. For the streaming pipeline the removal happens inside `ytdlPipeStreamer.Close` (after yt-dlp has been reaped, before `Close` returns), so quitting cliamp mid-track cannot leave a copy behind.
- A `cookies_file` that can't be read or copied is an **error before yt-dlp is spawned**. yt-dlp otherwise treats a missing path as an empty jar, runs unauthenticated, and creates the file on exit; a typo would silently degrade to bot-wall errors.
- NetEase reads its account cookies straight from the file, so account API calls need no yt-dlp in that mode.
- Provider `Config`s take one `Cookies ytdlcookies.Source`; `config` exposes `CookieSource()` per section.
- Docs for all four providers, `docs/configuration.md` (new "Cookies for yt-dlp Providers" section), `config.toml.example`, and the site provider notes are updated.

Not in scope: `cliamp setup` still offers browsers only.

#### Test plan

- [x] `make check` (gofmt, vet, `go test ./...`) on macOS; `GOOS=windows go vet` on touched packages (the CI matrix runs tests on Windows)
- [x] `ytdlcookies`: private copy contents/cleanup, read-only source, missing / directory / unreadable file -> error with nil args, temp-dir failure -> error
- [x] `resolve`: fake yt-dlp captures `--cookies <copy>` and the copy is gone after exit; a bad file fails before yt-dlp runs
- [x] `config`: all four sections parse `cookies_file`, `~` expansion, blank value reads as unset, `IsSetOrFallback` counts a file as credentials
- [x] `netease`: account API via cookies file with no yt-dlp on `PATH`; file without NetEase cookies and missing file both fail with `netease:` context
- [x] `player`: `Close` removes the cookie copy after yt-dlp exit and before returning (unit test on the streamer, plus a fixture-driven `decodeYTDLPipe` -> `Close` check that no `cliamp-cookies-*` file remains)
- [x] Verified against yt-dlp 2026.08.19: read-only `--cookies` -> `PermissionError` exit 1; writable -> rewritten on exit; missing -> exit 0, empty jar, file created; `~` in `--cookies` expanded by yt-dlp
- [ ] Live authenticated playback with a real exported `cookies.txt` (not exercised)

Generated with [Devin](https://devin.ai)


gjermundgaraba:
Dropped: the e2e harness handles cookies through yt-dlp's own config file (/etc/yt-dlp.conf in the SUT), so no cliamp change is needed.

PR #494, opened 2026-09-13
Preserve symlinks when saving files

rholak:
## Summary

Saving a symlinked config or playlist currently replaces the link with a
regular file. This PR keeps the link intact and updates its target instead.

Playlist saves and restores now use the shared atomic-write helper rather
than their own write-and-rename code. The helper leaves playlist directory
permissions alone and respects umask for new files. Existing file permissions
are never broadened, and dangling symlinks return an error.

Includes regression tests for symlinks and permission handling.

## Screenshots / video

N/A - this is related to the file-saving backend; no UI changes

## How to test

1. Run `make check`.
2. Run `go test -race ./internal/fileutil ./external/local ./config`.
3. For manual verification, build and run cliamp from this branch:
   - Back up your config and a playlist.
   - Move those files elsewhere and replace their original paths with symlinks.
   - Change a setting that cliamp saves, and add a track to the playlist.
   - Confirm both symlinks remain intact and the target files contain the changes.

## Checklist

- [x] `make check` passes
      — unchanged in this filesystem fix.
      

## Summary by CodeRabbit

- **Bug Fixes**
  - Playlist updates now preserve symlinks instead of replacing them.
  - Playlist and file restores retain directory permissions and correctly update symlink targets.
  - Atomic file writes now handle symlinked paths, missing targets, restrictive permissions, and umask settings more reliably.
- **Tests**
  - Added coverage for symlink preservation, directory permissions, path resolution, and umask behavior.


rholak:
CI passed on my fork for the submitted commit (c00dcec9):
https://github.com/codefriendly/cliamp/actions/runs/34780089142

internal/fileutil/atomic.go
```diff
@@ -1,27 +1,63 @@
...
+// directory's permissions. The parent directory must already exist.
+func WriteFileAtomicInExistingDir(path string, data []byte, perm os.FileMode) error {
+	return replaceFileAtomic(path, data, perm, false)
+}
+
+func replaceFileAtomic(path string, data []byte, perm os.FileMode, secureDir bool) (err error) {
+	originalDir := filepath.Dir(path)
+	info, err := os.Lstat(path)
+	if err == nil && info.Mode()&os.ModeSymlink != 0 {
+		path, err = filepath.EvalSymlinks(path)
```
gjermundgaraba: a dangling link now fails the save where main silently replaced the link with a regular file, and the theme, audio-device and resume-state callers discard that error (`ui/model/update.go:929,1063`, `internal/resume/resume.go:52`), so a dotfiles link created before its target exists means those settings quietly never persist. Please resolve the link with `os.Readlink` to its final target and create the file there when it is missing, as `open(2)` does, keeping the failure for a target directory that is missing or read-only. Verified on this head: that variant passes every new test except `TestWriteFileAtomicRejectsDanglingSymlink`, which would then assert the opposite of the wanted behavior.

internal/fileutil/atomic.go
```diff
@@ -1,27 +1,63 @@
...
+
+// WriteFileAtomicInExistingDir replaces path without changing its parent
+// directory's permissions. The parent directory must already exist.
+func WriteFileAtomicInExistingDir(path string, data []byte, perm os.FileMode) error {
+	return replaceFileAtomic(path, data, perm, false)
+}
+
+func replaceFileAtomic(path string, data []byte, perm os.FileMode, secureDir bool) (err error) {
+	originalDir := filepath.Dir(path)
+	info, err := os.Lstat(path)
```
gjermundgaraba: the destination is now inspected before the secure-directory step runs, so a config directory that lacks the execute bit fails with `inspect destination: lstat …: permission denied`, where main repaired the directory to `0700` and then wrote. Verified on this head and on the merge-base with the directory at `0600`: head returns the error and the file keeps `old`, base succeeds and leaves the directory at `0700`.

gjermundgaraba (changes requested):

internal/fileutil/atomic.go
rholak: I added the `resolveAtomicDestination` function to follow symlink chains with os.Readlink, so it can create the final target file if it’s missing.
The symlink stays intact, and missing or read-only target directories still cause the save to fail.

I replaced the dangling-link rejection test with target-creation coverage and added tests for chained links and permission failures. make check passes locally.

internal/fileutil/atomic.go
rholak: Fixed by moving destination resolution after the existing directory-hardening step. `resolveAtomicDestination` now runs only after the original directory has been restored to 0700.

I added a regression test that starts with a directory at 0600, checks that the save succeeds, updates the file, and leaves the directory at 0700. make check passes locally.

gjermundgaraba (approved):

PR #507, opened 2026-09-14
fix(resolve): route M3U/PLS/feed client through the SOCKS5-aware transport

Atarit0:
resolve.go builds its own `httpClient`/`sniffClient` (used for M3U, PLS and RSS feed resolution) on top of `http.DefaultTransport` directly. `http.DefaultTransport` does not understand `socks5(h)://` proxy URLs — the same limitation #432 fixed for the audio streaming client in `internal/httpclient`. With `HTTPS_PROXY`/`HTTP_PROXY` set to a SOCKS5 proxy, playlist/feed resolution silently bypasses it and dials the target directly.

Behind a corporate SOCKS5 proxy with TLS-inspecting egress this surfaces as a confusing `x509: certificate signed by unknown authority` error on the direct connection, instead of resolving through the proxy like the rest of the app.

**Fix:** reuse `httpclient.Streaming.Transport` (the SOCKS5-aware RoundTripper already used for streaming) as the base transport for `httpClient` and `sniffClient`, keeping their existing per-purpose timeouts and the `uaTransport` UA wrapper unchanged.

**Testing:** verified against a real local SOCKS5 proxy. Without the fix, `resolveM3U` bypasses the proxy entirely. With the fix, resolving `https://radio.cliamp.stream/streams.m3u` through `HTTPS_PROXY=socks5h://127.0.0.1:PORT` succeeds and returns the expected track list, matching direct (no-proxy) behavior with no regression.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved feed and playlist resolution when using SOCKS proxy URLs.
  * Improved URL classification and stream detection through configured transport settings, including SOCKS proxy support.

* **Documentation**
  * Clarified transport behavior for feed and playlist connections when SOCKS proxy URLs are used.


gjermundgaraba:
thanks! the SOCKS5 routing part looks like a straightforward extension of #432.

The PR also enables every Go cipher suite, including `tls.InsecureCipherSuites()` (RSA key exchange, 3DES, RC4), for all streaming TLS connections. That is not mentioned in the description and is unrelated to the proxy fix: resolve already used Go's default cipher list through `http.DefaultTransport`, so the routing change alone does not need it. It is also a security trade-off that applies with or without a proxy.

Could you drop that block from this PR and open it separately, with the servers it is meant to help? Then this one stays the two-file proxy fix its title describes and I can review it as such.

Atarit0:
Good catch, thanks. That cipher block came from an unrelated local change of mine that leaked into this commit — it has nothing to do with the SOCKS5 routing fix, you're right that resolve.go already worked fine on Go's default cipher list via http.DefaultTransport. Force-pushed a cleaned-up commit that drops it; this PR is now exactly the two-file proxy fix from the title/description, nothing else.

I'll open the legacy-cipher change separately if/when I have a clean writeup with the actual servers behind it.

Atarit0:
Closing this one. Re-tested more carefully after your catch on the cipher block: it turns out net/http.Transport already dials socks5/socks5h proxies natively (verified against a real SOCKS5 proxy, with and without this patch — both resolve the M3U fine), so this fix was not actually fixing anything. My original diagnosis on the corporate cert error was wrong; that turned out to be a TLS-inspection/CA-trust issue on that network, unrelated to proxy routing or to cliamp at all.

Sorry for the noise, and thanks for the careful review and for cliamp itself, it is genuinely great software.

PR #524, opened 2026-09-19
fix(lua): don't block the UI loop on a busy visualizer plugin

gaurabxkc:
## Summary

A Lua plugin that is both the active visualizer and an event handler can freeze the whole program.

`invokeHook` holds the plugin's Lua lock in its own goroutine while the hook calls back into the UI (`cliamp.message`, player controls), and those calls block until the UI loop consumes them. `RenderVis` runs on that same UI loop and waited for the lock with `Lock()`, so the two waited on each other forever.

- `RenderVis` now uses `TryLock` and returns the previous frame when the plugin is busy, which is already how a slow render is handled.
- `InitVis` / `DestroyVis` wait at most 100 ms and skip the optional init/destroy call rather than hang.

## Screenshots / video

Not applicable: no visible change except that the freeze no longer happens.

## How to test

1. `go test ./luaplugin/ -run TestRenderVisDoesNotBlockOnBusyPlugin`: the test holds the lock the way `invokeHook` does. It times out on `main` and passes with this change.
2. `make check`

## Checklist

- [x] `make check` passes

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented visualizer operations from hanging when a plugin is busy.
  * Visualizer rendering now returns the previous frame when the plugin is temporarily unavailable.
  * Initialization and cleanup exit promptly if the plugin cannot be locked within a short timeout.


gaurabxkc:
A concrete repro, since this one is easy to read as theoretical and it is not. It hangs cliamp completely, with no way out but killing the process.

Any plugin that is both a visualizer and has a key binding will do it. A pomodoro plugin is the natural example: it draws the countdown as the visualizer, and binds a key to start a session.

1. Install a visualizer plugin that binds a key, and have the key handler call back into the UI, via `cliamp.message(...)` or a player control.
2. Switch the visualizer to that plugin, so `RenderVis` is running every frame.
3. Press the key.

The key handler runs in its own goroutine holding the plugin's Lua lock, and blocks on the UI loop consuming its `cliamp.message`. The UI loop is inside `RenderVis`, waiting on that same lock with `Lock()`. Neither side can move, and the terminal is frozen until the process is killed.

I hit this tonight while testing an unrelated feature against `main`, spent a while assuming my own key dispatch was at fault, and only recognised it after reading `RenderVis` again. The `TryLock` in this PR fixes it: a busy plugin simply holds its previous frame for a tick, which is already how a slow render is handled.

Happy to rebase if that helps.

PR #525, opened 2026-09-19
fix(lua): let plugins bind uppercase keys

gaurabxkc:
## Summary

Plugins cannot bind uppercase keys: the binding is registered but never fires.

`p:bind()` stores the key under `normalizeKey()`, which lowercased it, but `EmitKey` looked the pressed key up as-is. Bubbletea reports shift+h as `"H"`, so `p:bind("H", ...)` was stored as `"h"` and never matched. Lowercasing also made `"H"` collide with the core's reserved `"h"`, so the bind could be refused outright.

- `normalizeKey` keeps the case of a bare single character, since `h` and `H` are different keys to the core keymap too.
- Modified and named keys (`Ctrl+E`, `Left`) are still lowercased.
- `EmitKey` normalizes the same way, so registration and dispatch agree.

## Screenshots / video

Not applicable: no UI changes.

## How to test

1. `go test ./luaplugin/ -run TestPluginBind`: `TestPluginBindAndEmit` now binds `"X"` and checks that `"X"` fires and `"x"` does not (it fails on `main`). `TestPluginBindNormalizesModifiedKeys` checks that `Ctrl+E` still matches `ctrl+e`.
2. `make check`

## Checklist

- [x] `make check` passes

## Summary by CodeRabbit

- **Bug Fixes**
  - Key bindings now dispatch reliably when emitted key combinations use different capitalization, such as `Ctrl+E` and `ctrl+e`.
  - Bare single-character keys preserve case, so bindings for `X` and `x` remain distinct.


tahadx (approved):
## Review — approve

**What this fixes:** `p:bind("H", fn)` registered but never fired. Root cause is a registration/dispatch mismatch in `luaplugin/api_keymap.go`: `bind()` stored keys via `normalizeKey()` (which lowercased everything), while `EmitKey` looked the pressed key up verbatim. Bubbletea reports Shift+h as `"H"`, so a binding stored as `"h"` never matched an incoming `"H"`. Worse, lowercasing made `"H"` collide with the core-reserved `"h"`, so the bind could be refused outright.

**The change (2 files, +48/-4):**
- `normalizeKey` now preserves case for a bare single rune (`h` vs `H` stay distinct, matching the core keymap where they are different commands) and still lowercases modified/named keys (`Ctrl+E` → `ctrl+e`, `Left` → `left`), matching the core registry convention. Whitespace trim retained; single-rune check is rune-aware.
- `EmitKey` normalizes the incoming key the same way before dispatch, so registration and dispatch agree. `bind`/`unbind` already normalized.
- Tests: `TestPluginBindAndEmit` now asserts `"X"` fires only on `"X"` (fails on main); new `TestPluginBindNormalizesModifiedKeys` asserts `Ctrl+E` matches `ctrl+e`.

**Verification (local, on `fix/lua-uppercase-keys` @ 5627344):**
- `go test ./luaplugin/ -run TestPluginBind -v -count=1` — all 5 bind tests PASS.
- `go test ./luaplugin/... -count=1` — PASS.
- `go vet ./luaplugin/` — clean.
- CI: all checks SUCCESS; mergeable CLEAN.

**Risk:** very low. Isolated to Lua keymap path; no core keymap changes; no docs/site update needed (no user-facing doc change). One follow-up note: pairs naturally with #528 (reserved-key guard runs during `New()` before `SetReservedKeys`) — reviewing that next.

PR #528, opened 2026-09-19
fix(lua): apply reserved keys while plugins load

gaurabxkc:
## Summary

Plugins can bind keys the core owns, and those bindings silently never fire.

Plugins call `p:bind()` from their top-level chunk, which runs inside `luaplugin.New()`. `main.go` only calls `SetReservedKeys` after `New()` returns, so the reserved-key guard in `bind()` sees an empty set during load and accepts the binding. The core then consumes that key before the plugin dispatcher is reached, so the binding never fires and the plugin author gets no warning.

- `luaplugin.SetDefaultReservedKeys` records the reserved set before `New()`; each new `Manager` starts from a copy of it.
- `main.go` calls it with `model.ReservedKeys()` just before `New()`.
- `SetReservedKeys` still works on an existing `Manager`, and a nil default keeps today's behaviour.

With this, a plugin binding a core key gets `false` plus the existing "refusing to bind … reserved by cliamp core" warning at load time.

## Screenshots / video

Not applicable: no UI changes.

## How to test

1. `go test ./luaplugin/ -run Reserved`. `TestDefaultReservedKeysApplyDuringLoad` loads a plugin that binds a reserved key through `New()`. It fails on `main` and passes here. The test clears `CLIAMP_CONFIG_DIR` and `XDG_CONFIG_HOME` and asserts only the test plugin loaded.
2. `make check`

## Checklist

- [x] `make check` passes

## Summary by CodeRabbit

- **Bug Fixes**
  - Reserved Lua binding keys are now recognized before plugins load, preventing plugins from binding keys owned by the core.
  - Conflicting bindings are rejected during plugin initialization with the expected warning.

- **New Features**
  - Applications can configure core-reserved binding keys before creating the plugin manager.


gaurabxkc:
This may be more relevant now than when I opened it.

On #547 you wrote that if uppercase is ever opened to plugins, "cliamp should always be the first in line to own a keybind". That is what this PR fixes, independent of the uppercase question.

Right now the core is *not* first in line. Plugins call `p:bind()` from their top-level chunk, which runs inside `luaplugin.New()`, and `main.go` only calls `SetReservedKeys` after `New()` returns. So during load the reserved set is empty, the guard in `bind()` accepts anything, and the plugin's binding is recorded. The core still consumes the key first at dispatch, so the binding never fires: the plugin author gets no warning, and the user gets a key that looks bound and does nothing.

This PR records the reserved set before `New()`, so `bind()` can refuse a core key at the moment it is asked and say so in the plugin log.

That ordering is the precondition for whatever you decide about uppercase later. With it in place, opening a letter class to plugins is safe by construction, because the core owns its keys before any plugin runs. Without it, the guarantee does not hold for lowercase either.

Happy to rebase if it has drifted.

